PR #1, opened 2023-04-01
feat(sdk): add sdk initially

chokobole:
SDK contains mainly 2 features, bridge and transaction fee estimation. Some of legacy features are removed from optimism SDK, such as, CTC and SCC related. Encoding and hashing message scheme is also adapted to fit in our case.


packages/sdk/LICENSE
```diff
@@ -0,0 +1,22 @@
+(The MIT License)
+
+Copyright 2023 Kanvas
```
Pangssu, line 3: I think it should be `WEMIX Kanvas`

PR #4, opened 2023-04-03
fix(validator): avoid infinite polling

chokobole:
validator hits cpu usage 100% because select statement in Validator.loop() doesn't do what we expect. This should sleep and wake up at every tick.


Pangssu (approved):
lgtm

Issue #5, opened 2023-04-03
batcher doesn't send batch again when it's failed to send.


chokobole:
### Issue Type

Bug

### Kanvas Version

0.1.0

### OS Platform and Distribution

Linux Ubuntu 22.04

### Go version

go1.19.3 linux/amd64

### Node version

v16.14.2

### Current Behaviour?

One thing you should know why it reorged around in the morning at L2 block `118209`, this is because we had a mistake configuring `ProposerWindowSize` with different values to `vanilla` node and `proposer/validator` nodes.

- `proposer` and `validator`: 300
- `vanilla`: 3600

This leads l2 reorgs and vanilla doesn't follow the head of proposer. The core reason why it happened this is because batcher didn't resend the batch again and batcher kept saving batch and channel rejected because channel was already closed.

### Relevant log output

See timestamp differences, approximately 1 hr. 

- [batch_{i - 1}](https://sepolia.etherscan.io/tx/0x606d61232ceb5e01e697f9643b121588228cd68a32cd8da4d2f982cf2e05ce78) at `1680480852`
- [batch_i](https://sepolia.etherscan.io/tx/0x133c5dead00daf5404b424ca86a0c6c3eba4f7c3fdf856994c7a5dfac2414536) at `1680477264`

chokobole:
@Pangssu Could you tackle down this problem?

PR #6, opened 2023-04-03
fix(batcher): fix issue of not retrying failed batch tx before sending

Pangssu:
This commit fixes the issue of not retrying when encountering errors during the creation of batch submission transaction.

Resolves: #5


chokobole (approved):
LGTM

PR #16, opened 2023-04-04
chore: inject git tag as a version

Pangssu:
The Makefile is modified to set the versions of the binaries using git tag. 
The versions are managed using git tag and GitHub release, and the version of the binaries is set to the tag associated with the commit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
+VERSION := $(shell git describe --tags --abbrev=0 --match v* 2> /dev/null || git rev-parse --short HEAD)
 GIT_COMMIT := $(shell git rev-parse HEAD)
 GIT_DATE := $(shell git show -s --format='%ct')
-VERSION := v0.1.0
 
 LD_FLAGS_ARGS +=-X main.Version=$(VERSION)
 LD_FLAGS_ARGS +=-X main.GitCommit=$(GIT_COMMIT)


Pangssu:
Should this commit type be `build`? @chokobole

chokobole:
@Pangssu Please do it for next time, at this moment, i'll accept this commit type.

Makefile
```diff
@@ -1,6 +1,6 @@
+VERSION := $(shell git describe --exact-match --tags 2> /dev/null || git rev-parse --short HEAD)
```
chokobole, line 1: This gives us tag only when `HEAD` is tagged.

```shell
> git describe --exact-match --tags
fatal: no tag exactly matches 'd12c2add1cf474f707e2edf7ce383ac7caf38e11'
> git tag @wemixkanvas/a@0.1.1
> git describe --exact-match --tags
@wemixkanvas/a@0.1.1
```

Please update this with below, this gives us the most recent tag that starts with `v`. Because we mix tagging with npm `@wemixkanvas/...`, and I don't want component to be affected by them

```diff
-VERSION := $(shell git describe --exact-match --tags 2> /dev/null || git rev-parse --short HEAD)
+VERSION := $(shell git describe --tags --abbrev=0 --match v* 2> /dev/null || git rev-parse --short HEAD)
```

chokobole: I also checked this working well inside the docker container, too.

Pangssu:
@chokobole What do you think about standardizing the version format of components? Currently, the batcher and validator use the `{version}-{commit}-{timestamp}` format, while the node uses the `{version}-{commit 8 chars}-{timestamp}-{meta}` format. I propose to standardize the format to `{version}-{commit 8 chars}-{timestamp}`.

chokobole:
Good, could you please send another PR for this?

Pangssu:
Okay, I'll leave it as a new PR.

chokobole (approved):
LGTM

Issue #11, opened 2023-04-03
Support debugging by attaching go process which is run by docker compose


chokobole:
### Issue Type

Feature Request

### Kroma Version

0.1.0

### OS Platform and Distribution

Linux Ubuntu 22.04

### Go version

go1.19.3 linux/amd64

### Current Behaviour?

In `ops-devnet/docker-compose.yaml`, set `pid` to `host` on `kanvas-node`.

```yaml
kanvas-node:
  pid: host
```

And then set `ptrace_scope` to `0`

```shell
> sudo su -
> echo 0 >  /proc/sys/kernel/yama/ptrace_scope
```

And then if I try to debug by attaching go process is rejected with this message.

```shell
Failed to attach: Could not attach to pid 39997: current user does not own the process.
```

### Which feature do you want to expect?

I expect the process is launched with `uid:gid` so that we are able to debug. Currently, we should rely on the log message, which is very difficult when we consider that the logs are emitted so long.

PR #23, opened 2023-04-11
chore: support debugging devnet process

chokobole:
This commit helps user to debug a process which is run by docker compose. Currently, this only works on Linux. There should be another way for macOS users.

Resolves: #11


specs/meta/devnet.md
```diff
@@ -51,3 +52,10 @@ The batch submitter uses the account below to submit batches to L1:
 
 - Address: `0xde3829a23df1479438622a08a116e8eb3f620bb5`
 - Private key: `bf7604d9d3a1c7748642b1b7b05c2bd219c9faa91458b370f85e5a40f3b03af7`
+
+## Debugging vscode
```
Pangssu, line 56: VS Code

Dockerfile
```diff
@@ -16,19 +16,55 @@ RUN make build
...
 FROM alpine:3.17 as kanvas-node
 COPY --from=builder /app/bin/kanvas-node /usr/local/bin
+
+RUN addgroup user && \
+    adduser -G user -s /bin/sh -h /home/user -D user && \
+    mkdir /kanvas_log/ && \
+    chown user:user /kanvas_log
+
+USER user
+WORKDIR /home/user/
```
Pangssu, line 26: I think it's better to separate the stage for adding this.
Is this possible?

```suggestion
FROM alpine:3.17 as runner
RUN addgroup user && \
    adduser -G user -s /bin/sh -h /home/user -D user && \
    mkdir /kanvas_log/ && \
    chown user:user /kanvas_log
USER user
WORKDIR /home/user/

# Node
FROM runner as kanvas-node
COPY --from=builder /app/bin/kanvas-node /usr/local/bin
...

File: Dockerfile [hunk 3]
```diff
@@ -16,19 +16,55 @@ RUN make build
...
 CMD ["kanvas-stateviz"]
 
 # Batcher
 FROM alpine:3.17 as kanvas-batcher
 COPY --from=builder /app/bin/kanvas-batcher /usr/local/bin
+
+RUN addgroup user && \
+    adduser -G user -s /bin/sh -h /home/user -D user && \
+    mkdir /kanvas_log/ && \
+    chown user:user /kanvas_log
```
Pangssu, line 51: FYI, batcher and validator do not required a `kanvas_log` file.

Dockerfile
Pangssu: This code will not work and is for reference purposes only.

Dockerfile
chokobole: Oh thanks i'll remove the lines on `batcher` and `validator`

PR #25, opened 2023-04-13
feat!: include next block hash

chokobole:
Between the asserter and challenger, the source output root and destination
output root are ultimately determined, but the next block hash required
for ZK fault proof cannot currently be obtained from them. This commit resolves
this issue by having both participants agree to include the next block hash
in the source output root.

Related: https://github.com/wemixkanvas/kanvas/issues/17


chokobole:
We need to update `l2GenesisBlueTimeOffset` in  `packages/contracts/deploy-config/easel.json` after network go stable.

chokobole:
I talked with @kangsorang, the blue hardfork should be aligned with `SUBMISSION_INTERVAL` so that all the outputs within the same range can be calculated with same hashing mechanism.

Issue #10, opened 2023-04-03
TestWithdrawals is failed on CI


chokobole:
### Issue Type

Bug

### Kanvas Version

0.1.0

### OS Platform and Distribution

Linux Ubuntu 22.04

### Go version

go1.19.3 linux/amd64

### Node version

v16.14.2

### Current Behaviour?

`TestWithdrawals` test is failed on ci.

### Standalone code to reproduce the issue

```shell
go test ./e2e/ (on ci)
```

### Relevant log output

```shell
2023-03-30T05:42:10.8878967Z     system_test.go:1181: 
2023-03-30T05:42:10.8879407Z         	Error Trace:	/home/runner/work/kanvas/kanvas/e2e/system_test.go:1181
2023-03-30T05:42:10.8881352Z         	Error:      	Expected nil, but got: &rpc.jsonError{Code:3, Message:"execution reverted: KanvasPortal: proven withdrawal finalization period has not elapsed", Data:"0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000434b616e766173506f7274616c3a2070726f76656e207769746864726177616c2066696e616c697a6174696f6e20706572696f6420686173206e6f7420656c61707365640000000000000000000000000000000000000000000000000000000000"}
2023-03-30T05:42:10.8881622Z         	Test:       	TestWithdrawals
```

See https://github.com/wemixkanvas/old-kanvas/actions/runs/4560939997/jobs/8046359187

chokobole:
Reopen the issue if it happens again. Seems https://github.com/wemixkanvas/kanvas/pull/26 resolved the issue

Issue #9, opened 2023-04-03
TestBatchInLastPossibleBlocks is failed on CI


chokobole:
### Issue Type

None

### Kanvas Version

0.1.0

### OS Platform and Distribution

Linux Ubuntu 22.04

### Go version

go1.19.3 linux/amd64

### Node version

v16.14.2

### Current Behaviour?

`TestBatchInLastPossibleBlocks` fails time to time. Need to check why it sometimes failed. I am suspicious around here.

```shell
2023-03-29T11:12:12.1902089Z ERROR[03-29|11:03:51.317] Missing block in the middle, aiming genesis number=18,446,744,073,709,551,615 hash=000000..000000
```

### Standalone code to reproduce the issue

```shell
go test ./e2e/actions (on ci)
```

### Relevant log output

See https://pipelines.actions.githubusercontent.com/serviceHosts/21e583e2-d6a0-45c6-8467-1553461385db/_apis/pipelines/1/runs/286/signedlogcontent/2?urlExpires=2023-03-29T11%3A32%3A49.0624750Z&urlSigningMethod=HMACV1&urlSignature=Q1wcY7o18Um9yB9n0QTBlRM8lOrW%2B%2F%2BWxvmn7SHkRUg%3D

chokobole:
Found more example, https://github.com/wemixkanvas/kanvas/actions/runs/4669333116/jobs/8267631293?pr=24

chokobole:
I'll reopen the issue if it happens again. Seems https://github.com/wemixkanvas/kanvas/pull/26 resolved the issue

PR #26, opened 2023-04-14
ci: add disable parallel option in e2e test

ray-ijtt:
Exclude unpredictable behavior due to the parallel environment during e2e test operation. 

Related #9, #10


chokobole (approved):
LGTM

PR #32, opened 2023-04-17
feat: include next block hash

chokobole:

Issue #17, opened 2023-04-05
Include next block hash in output root


chokobole:
### Issue Type

Feature Request

### Kanvas Version

0.1.1

### OS Platform and Distribution

Linux Ubuntu 22.04

### Go version

go1.19.3 linux/amd64

### Node version

v16.14.2

### Changes

**NOTE: This is a temporary solution before Proto Dank Sharding.**

To resolve the dispute, `Colosseum` contract needs to verify ZK fault proof in the end. A ZK fault proof proves that state transition `S` to `S'` is valid. Currently, a proof isn't constrained with public input. What this means is that any valid proof can be verified, so you can win the challenge all the times if you manage to generate a proof.

To confine a specific state transition, it needs a public input such as transaction root. At this moment, there's no such info going on Layer 1. I am thinking that the best approach to do this is to include `next_block_hash` in `output_root` calculation. 

```
version = 1
output_root = hash(version || state root || withdrawal storage root || block hash || next block hash)
```

Output submission will take the same cost, but there's a small increment for withdrawal. But considering output submission happens more often, this gets reasonable. This also breaks a little compatibility from `optimism`, thus we need to take care when adopting optimism codes.

Maybe we can create `PublicInputOracle.sol` that behaves almost same as `L2OutputOracle.sol` but gives us a public input for a specific block number. But this requires for batcher too much cost, because it should be done per batch, I believe.

When applying an above strategy, a pair of output roots are determined during interaction, we can safely get the data from `next block hash` of source output root.

Here's an example: let's say you have a pair of output roots between block 9 and block 10, and you know the `OutputRoot9` and `OutputRoot10` of the asserter. You can take the `BlockHash10` from `OutputRoot9` and use it to verify the ZK fault proof. 

### Security Consideration

In the example above, when `OutputRoot9` and `OutputRoot10` are remained, there is no way to verify the `next block hash` in `OutputRoot10`. Attacks utilizing this vulnerability must be prevented by relying on the [Security Council](#18).

### Tasks

- [x] `validator` node submits updated output root. (https://github.com/wemixkanvas/kanvas/pull/32)
- [x] `L2OutputOracle` stores output root and validates differently per version. (https://github.com/wemixkanvas/kanvas/pull/25)
- [x] `Colosseum` obtains public input for ZK fault proof from `OutputRootProof`. (https://github.com/kroma-network/kroma/pull/38)
[...]

PR #38, opened 2023-04-24
feat: prepare checking public input

chokobole:
Version 1 output root contains block hash since #32. Now the Colosseum
contract compute public input based on block hash and use it to verify
ZK fault proof.                                                                                                                                                 

Related: #17


Pangssu (approved):
lgtm

PR #41, opened 2023-04-26
chore: sort import statements

0xHansLee:
There are some unsorted import statements, so I sorted them using [`goimports-reviser`](https://github.com/incu6us/goimports-reviser).
```bash
> goimports-reviser -rm-unused ./...
```
Currently, `goimports` do not sort import statements consistently (it is well-known issue: https://github.com/golang/go/issues/20818).

There are many other alternatives, and I tried to use `goimports-reviser` since it works well as we desired and they also have linter for sorting import statements (we can add linter to CI in the future as well). 

FYI, you can configure `goimports-reviser` in IntelliJ as described [here](https://github.com/incu6us/goimports-reviser#example-to-configure-it-with-jetbrains-ides-via-file-watcher-plugin).

### DIFF
--- a/components/node/eth/account_proof.go
+++ b/components/node/eth/account_proof.go
@@ -14,7 +14,6 @@ import (
 	"github.com/ethereum/go-ethereum/ethdb/memorydb"
 	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/ethereum/go-ethereum/trie"
-
 	zktrie "github.com/kroma-network/zktrie/trie"
 	zkt "github.com/kroma-network/zktrie/types"
 )
--- a/components/node/withdrawals/proof.go
+++ b/components/node/withdrawals/proof.go
@@ -10,7 +10,6 @@ import (
 	"github.com/ethereum/go-ethereum/crypto/poseidon"
 	"github.com/ethereum/go-ethereum/ethclient/gethclient"
 	"github.com/ethereum/go-ethereum/trie"
-
 	zktrie "github.com/kroma-network/zktrie/trie"
 	zkt "github.com/kroma-network/zktrie/types"
 )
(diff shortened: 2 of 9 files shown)


chokobole (approved):
LGTM

components/node/eth/account_proof.go
```diff
@@ -14,7 +14,6 @@ import (
 	"github.com/ethereum/go-ethereum/ethdb/memorydb"
 	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/ethereum/go-ethereum/trie"
-
```
seolaoh, line 17: I think this line should not be removed, isn't it?

components/node/withdrawals/proof.go
```diff
@@ -10,7 +10,6 @@ import (
 	"github.com/ethereum/go-ethereum/crypto/poseidon"
 	"github.com/ethereum/go-ethereum/ethclient/gethclient"
 	"github.com/ethereum/go-ethereum/trie"
-
```
seolaoh, line 13: ditto

components/node/eth/account_proof.go
0xHansLee: Yes, it has been removed in the new version. This is the old version

seolaoh: I mistook our organization's package for the current package😅 Thank you!

seolaoh (approved):
LGTM

0xHansLee:
Let me share how to set `goimports-reviser` formatter on save in VS code. 
(I'm not sure this is the best way, but I want to share how to figure it out. If you have any better idea, please leave your tips for everyone.)

I used [`run on save`](https://marketplace.visualstudio.com/items?itemName=emeraldwalk.RunOnSave) extenstion to run `goimports-reviser` command on save.
This is the configuration added to `settings.json`
```json
{
    // ...
    "emeraldwalk.runonsave": {
        "commands": [
            {
                "match": ".*",
                "cmd": "$HOME/go/bin/goimports-reviser -rm-unused ${file}"
            },
        ]
    }
    // ...
}
```

PR #42, opened 2023-04-27
feat(validator): apply op-proposer v1.0.3 to Kroma

seolaoh:
# Changelog

## v1.0.2
- https://github.com/ethereum-optimism/optimism/pull/5258
- https://github.com/ethereum-optimism/optimism/pull/5283
- https://github.com/ethereum-optimism/optimism/pull/5295

## Other changes
- https://github.com/ethereum-optimism/optimism/pull/5189
- https://github.com/ethereum-optimism/optimism/pull/5246

### DIFF
--- a/components/validator/cmd/main.go
+++ b/components/validator/cmd/main.go
@@ -23,9 +23,9 @@ func main() {
 	app := cli.NewApp()
 	app.Flags = flags.Flags
 	app.Version = fmt.Sprintf("%s-%s", Version, Meta)
-	app.Name = "validator"
-	app.Usage = "L2Output Submitter"
-	app.Description = "Service for generating and submitting L2 Output checkpoints to the L2OutputOracle contract"
+	app.Name = "kroma-validator"
+	app.Usage = "L2 Output Submitter and Challenger Service"
+	app.Description = "Service for generating and submitting L2 output checkpoints to the L2OutputOracle contract as an L2 Output Submitter, " + "detecting and correcting invalid L2 outputs as a Challenger to ensure the integrity of the L2 state."
 
 	app.Action = curryMain(Version)
 	err := app.Run(os.Args)
--- a/components/validator/metrics/metrics.go
+++ b/components/validator/metrics/metrics.go
@@ -1,0 +1,105 @@
+package metrics
+
+import (
+	"context"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/kroma-network/kroma/components/node/eth"
+	kmetrics "github.com/kroma-network/kroma/utils/service/metrics"
+	txmetrics "github.com/kroma-network/kroma/utils/service/txmgr/metrics"
+)
+
+const (
+	Namespace         = "kroma_validator"
+	L2OutputSubmitted = "submitted"
+)
+
+type Metricer interface {
+	RecordInfo(version string)
+	RecordUp()
+
+	// Records all L1 and L2 block events
+	kmetrics.RefMetricer
+
+	// Record Tx metrics
+	txmetrics.TxMetricer
+
+	RecordL2OutputSubmitted(l2ref eth.L2BlockRef)
+}
+
+type Metrics struct {
+	ns       string
+	registry *prometheus.Registry
+	factory  kmetrics.Factory
+
+	kmetrics.RefMetrics
+	txmetrics.TxMetrics
+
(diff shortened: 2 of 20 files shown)


components/validator/validator.go
```diff
@@ -135,24 +151,30 @@ func (v *Validator) loop() {
...
 	}
-	if err := v.SendTransaction(cCtx, tx); err != nil {
+
+	txCandidate := txmgr.TxCandidate{
+		TxData:   data,
+		To:       &v.cfg.L2OutputOracleAddr,
+		GasLimit: 0,
+	}
+
+	if err := v.sendTransaction(ctx, txCandidate); err != nil {
```
seolaoh, line 174: Here, there is a timeout of 10 minutes in case of Optimism before sending the transaction, but I didn't add it because it didn't exist originally in ours. If this part is missing, I will add it like below. Please check it.
```suggestion
	cCtx, cancel := context.WithTimeout(ctx, 10*time.Minute)
	defer cancel()
	if err := v.sendTransaction(ctx, txCandidate); err != nil {
```

components/validator/cmd/main.go
```diff
@@ -23,9 +23,9 @@ func main() {
 	app := cli.NewApp()
 	app.Flags = flags.Flags
 	app.Version = fmt.Sprintf("%s-%s", Version, Meta)
-	app.Name = "validator"
-	app.Usage = "L2Output Submitter"
-	app.Description = "Service for generating and submitting L2 Output checkpoints to the L2OutputOracle contract"
+	app.Name = "kroma-validator"
+	app.Usage = "L2 Output Submitter and Challenger"
+	app.Description = "Service for generating and submitting L2 output checkpoints to the L2OutputOracle contract and detecting invalid L2 outputs"
```
chokobole, line 28: 1. When seeing `components/batcher/cmd/main.go`, it might be better to add a suffix `Serivce` to `app.Usage`. What do you think?
2. For me, challenge not only detects invalid L2 outputs but also corrects it.

```suggestion
-app.Usage = "L2 Output Submitter and Challenger"
-app.Description = "Service for generating and submitting L2 output checkpoints to the L2OutputOracle contract and detecting invalid L2 outputs"
+app.Usage = "L2 Output Submitter and Challenger Service"
+app.Description = "Service for generating and submitting L2 output checkpoints to the L2OutputOracle contract and detecting and correcting invalid L2 outputs to ensure the integrity of the L2 state."
```

components/validator/metrics/metrics.go
```diff
@@ -0,0 +1,106 @@
...
+
+// RecordUp sets the up metric to 1.
+func (m *Metrics) RecordUp() {
+	prometheus.MustRegister()
+	m.Up.Set(1)
+}
+
+const (
+	L2OutputSubmitted = "submitted"
+)
```
chokobole, line 101: Could you move up and merge all the consts?

```go
const (
  Namespace = "kroma_validator"
  L2OutputSubmitted = "submitted"
)
```

components/validator/validator.go
chokobole: I think we need this in case there's no response from a server!

seolaoh: `sendTransaction`, which is called after the context declaration with 10 minutes timeout, calls `TxManager.Send`. At the beginning of `TxManager.Send`, there is a part to declare timeout context using `TxManager.Config.TxSendTimeout`.
So I thought that the 10 minute timeout before `sendTransaction` is unnecessary, wdyt?

components/validator/cmd/main.go
seolaoh: I changed `components/node/cmd/main.go` also to unify `app.Usage` and `app.Description`. Check it also plz.

components/validator/validator.go
seolaoh: I added some env variables to `ops-devnet/docker-compose.yml` to set `TxManager.Config.TxSendTimeout` 10 min.

utils/service/txmgr/cli.go
```diff
@@ -34,21 +34,6 @@ const (
...
-			"mnemonic. The mnemonic flag must also be set.",
-		EnvVar: "BATCHER_HD_PATH",
-	}
-	ValidatorHDPathFlag = cli.StringFlag{
-		Name: "validator-hd-path",
-		Usage: "DEPRECATED: The HD path used to derive the validator wallet from the " +
-			"mnemonic. The mnemonic flag must also be set.",
-		EnvVar: "VALIDATOR_HD_PATH",
-	}
-)
```
seolaoh, line 50: I removed some legacy flags like `BatcherHDPathFlag`, `ValidatorHDPathFlag` cause it is duplicated with `HDPath`.

chokobole (approved):
LGTM

Issue #44, opened 2023-04-28
After binding creation, "'TypesOutputRootProof' has been redeclared in this package" occurs.


Pangssu:
### Issue Type

Bug

### Kanvas Version

0.1.1

### OS Platform and Distribution

Windows 11

### Go version

go1.20.2

### Node version

v16.18.0

### Current Behaviour?

```shell
In commit ['1c8e695b'](https://github.com/kroma-network/kroma/commit/1c8e695b16ff03ef0ff34c913d02a7c9b897e041#diff-77450785d1dafe0fddac317efe5f5151f0cd2d57405a72bccf457145c67e24a7), the `proveFault` function in the `Colosseum.sol` was modified to take the `Types.OutputRootProof` structure as a parameter. `Types.OutputRootProof` structure is also used in `KromaPortal.sol`. As a result of this modification, the problem of duplicate generation of the `TypesOutputRootProof` structure within the same package occurs when the bindings are generated by `abigen`.
```

### Standalone code to reproduce the issue

```shell
make bindings
```

### Relevant log output

```shell
'TypesOutputRootProof' redeclared in this package
```

Pangssu:
solved

PR #45, opened 2023-04-28
feat: extract and generate Types struct bindings

Pangssu:
# Description
implements feature that the extraction of the struct declared in the Types.sol library used in the contract and creates a new binding, thus preventing duplication of the same struct across existing bindings.

I think this is experimental feature. If there is a possibility of problems arising from this feature, it is okay to boldly close the PR.

Resolves: #44


chokobole (approved):
LGTM

PR #46, opened 2023-04-28
chore: change git action runner to selfhost

ccrayz:
we will maintain `self-hosting` for the time being to improve the issue of failed CI due to limited pipeline performance


Pangssu (approved):
lgtm

PR #51, opened 2023-05-15
ci: change github action runner host

ccrayz:
I have been using self-hosting to utilize more computing resources than the default specifications of GitHub Actions when running actions. 
However, I now have the option to use `larger runners`, so I'm switching to that.

> Additionally, we would like to lock/freeze the nightly version.


chokobole (approved):
LGTM

PR #52, opened 2023-05-17
ci: change dev tag name

ccrayz:
change the policy to differentiate commits that are merged into the dev branch by tagging them in the docker image

also, the previously used release version is no longer available, so it is being replaced with a compatible alternative version.


chokobole (approved):
LGTM

PR #55, opened 2023-05-24
feat: submit genesis output at starting block number

seolaoh:
By changing to submit the genesis output at starting block number as the first output, enable the challenge for the next output. Genesis output can only be submitted by the trusted validator and cannot be challenged.
Checked all tests passed except for the e2e tests, we need to check the e2e tests after the validator component modified.


packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -310,10 +311,9 @@ contract L2OutputOracle is Initializable, Semver {
     }
 
     /**
-     * @notice Returns the number of outputs that have been submitted. Will revert if no outputs
-     *         have been submitted yet.
+     * @notice Returns the index of the latest submitted output.
      *
-     * @return The number of outputs that have been submitted.
```
Pangssu, line 316: Is there any reason the comment about `revert` was removed? It still seems to revert when `l2Outputs.length` is 0.

seolaoh: Oh I see, I will restore the comment. Thx!

Pangssu (approved):
LGTM

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -245,8 +245,7 @@ contract L2OutputOracle is Initializable, Semver {
     }
 
     /**
-     * @notice Returns an output by index. Exists because Solidity's array access will return a
-     *         tuple instead of a struct.
+     * @notice Returns an output by index. Reverts if output is not found at the given index.
```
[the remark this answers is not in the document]
Pangssu: This function will revert with `Index out of bounds` if the output corresponding to index does not exist.

components/validator/challenger.go
```diff
@@ -110,22 +111,22 @@ func (c *Challenger) GetInvalidOutputRange() (*OutputRange, error) {
...
+		waitingOutputNum := new(big.Int).Div(c.finalizationPeriodSeconds, submissionIntervalSeconds)
+		if latestOutputIndex.Cmp(waitingOutputNum) != 1 {
 			c.checkpoint = common.Big1
-		} else if waitingOutputs.Cmp(common.Big0) == 0 {
-			c.checkpoint = latestOutputIndex
-			return nil, nil
+			} else if waitingOutputNum.Cmp(common.Big0) == 0 {
+				c.checkpoint = latestOutputIndex
+				return nil, nil
 		} else {
```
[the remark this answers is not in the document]
Pangssu: Oh, I missed it.

components/validator/challenger.go
```diff
@@ -110,22 +111,22 @@ func (c *Challenger) GetInvalidOutputRange() (*OutputRange, error) {
...
+		waitingOutputNum := new(big.Int).Div(c.finalizationPeriodSeconds, submissionIntervalSeconds)
+		if latestOutputIndex.Cmp(waitingOutputNum) != 1 {
 			c.checkpoint = common.Big1
-		} else if waitingOutputs.Cmp(common.Big0) == 0 {
+		} else if waitingOutputNum.Cmp(common.Big0) == 0 {
 			c.checkpoint = latestOutputIndex
 			return nil, nil
 		} else {
-			c.checkpoint = new(big.Int).Mod(latestOutputIndex, waitingOutputs)
+			c.checkpoint = new(big.Int).Sub(nextOutputIndex, waitingOutputNum)
```
0xHansLee, line 129: It would be good to note this change: `Mod` to `Sub`

AFAIK, when `c.checkpoint` is nil and the `latestOutputIndex` is larger than `waitingOutputNum` (meaning that after the first finalization), the `c.checkpoint` should be the start point of the finalization window.

packages/contracts/contracts/test/Colosseum.t.sol
```diff
@@ -24,18 +22,21 @@ contract ColosseumTest is Colosseum_Initializer {
...
         // Roll to after the block number we'll submit
         warpToSubmitTime(nextBlockNumber);
         vm.prank(pool.nextValidator());
         oracle.submitL2Output(bytes32(nextBlockNumber), nextBlockNumber, 0, 0, minBond);
 
+        // Submit valid output
+        nextBlockNumber = oracle.nextBlockNumber();
+        warpToSubmitTime(nextBlockNumber);
+        vm.prank(pool.nextValidator());
+        oracle.submitL2Output(bytes32(nextBlockNumber), nextBlockNumber, 0, 0, minBond);
```
0xHansLee, line 36: Just question. Do you have any reason for adding this code block, submitting valid output?
IMHO, the genesis output is already been submitted (which is considered valid), there is no need to submit one more output.

seolaoh: That's because `INVALID_BLOCK_NUMBER` in `/packages/contracts/contracts/test/testdata/ColosseumTestData.sol` is `3000`. We use this value and related test data in Colosseum tests.
Otherwise, we can change `INVALID_BLOCK_NUMBER` to a number between 0 and 1800, and remove this code block. In this case we should change related test data (e.g. `PublicInput.number` in line 48 of `ColosseumTestData.sol`, or anywhere else?).

Pangssu:
Please rebase to the `feat/multiple-validator` branch and run `make bindings && make contracts-snapshot` from the project root.

packages/contracts/contracts/test/Colosseum.t.sol
0xHansLee: Ah, I got it. Thank you for your explanation 🙏

0xHansLee (approved):
lgtm

chokobole (approved):
LGTM

PR #59, opened 2023-05-26
feat!(challenger): change challenge handling to event-driven for multiple challenges

0xHansLee:
### Description

Challenge handling is changed from polling-based to event-driven way. As multiple challenges can be created simultaneously, the validator can properly handle multiple challenges at the same time.

To handle outputs in the past and within the finalization window, `scanPrevOutputs` has also been implemented.

### Tx nonce collision
`Validator` has two components: L2 output submitter and challenger.
Since each component sends transaction with the same account, should be very careful with nonce collisions.
To avoid it, `Validator` would have a channel for transaction, `txCandidatesChan`.
The created tx candidates will be transferred to `txCandidateChan` in each component, and `txmgr` will process the tx candidates sequentially. 

##

FYI, 
1. E2E tests have not been handled yet, just fixed type error. It will be handled after all the features are implemented.
2. Moved existing challenger functions down.

This PR has a slightly complicated structure, using several go routines to concurrently process multiple challenges. 
I'd appreciate you for your detailed review.


components/validator/challenger.go
```diff
@@ -65,191 +93,413 @@ func NewChallenger(ctx context.Context, cfg Config, l log.Logger) (*Challenger,
...
 		colosseumContract: colosseumContract,
+		colosseumABI:      colosseumABI,
 
 		submissionInterval:        submissionInterval,
 		finalizationPeriodSeconds: finalizationPeriodSeconds,
 		l2BlockTime:               l2BlockTime,
-	}, nil
-}
 
-func (c *Challenger) IsChallengeInProgress() (bool, error) {
```
0xHansLee, line 84: FYI, this code is not deleted actually, moved below.

components/validator/challenger.go
```diff
@@ -24,28 +32,48 @@ type ProofFetcher interface {
 }
 
 type Challenger struct {
-	log log.Logger
-	cfg Config
-	ctx context.Context
+	log  log.Logger
+	cfg  Config
+	ctx  context.Context
+	done chan struct{}
```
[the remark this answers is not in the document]
chokobole: By the way, who sends the data through `done`? I found no where.

components/validator/challenger.go
```diff
@@ -65,191 +93,413 @@ func NewChallenger(ctx context.Context, cfg Config, l log.Logger) (*Challenger,
...
+	return nil
 }
 
-type OutputRange struct {
-	OutputIndex *big.Int
-	StartBlock  uint64
-	EndBlock    uint64
+func (c *Challenger) Stop() {
+	c.log.Info("stop challenger")
+	close(c.done)
```
chokobole, line 149: If you don't send any data to `c.done`, please close the channel as early as possible.

components/validator/challenger.go
```diff
@@ -65,191 +93,413 @@ func NewChallenger(ctx context.Context, cfg Config, l log.Logger) (*Challenger,
...
+
+	// if challenge mode on, subscribe L2 output submission events
+	if !c.cfg.ChallengerDisabled {
+		c.wg.Add(1)
+		go c.subscribeL2OutputSubmitted()
+	}
+
+	// subscribe challenge creation events
+	c.wg.Add(1)
+	go c.subscribeChallengeCreated()
```
chokobole, line 142: Passing the current L2 block number as an explicit argument to functions like `scanPrevOutputs`, `subscribeL2OutputSubmitted`, and `subscribeChallengeCreated` can be a good approach. This allows for more flexibility and ensures that the functions are working with the correct and up-to-date L2 block number.

components/validator/challenger.go
```diff
@@ -65,191 +93,413 @@ func NewChallenger(ctx context.Context, cfg Config, l log.Logger) (*Challenger,
...
-		c.log.Info("challenge for genesis output is not allowed", "latestOutputIndex", latestOutputIndex)
-		return nil, nil
+// scanPrevOutputs scans all the previous outputs within the finalization window.
+// If there are invalid outputs, create challenge.
+// If there are challenges in progress, keep handling them.
+func (c *Challenger) scanPrevOutputs() {
+	defer c.wg.Done()
+
+	ctx, cancelCtx := context.WithTimeout(context.Background(), c.cfg.NetworkTimeout)
+	defer cancelCtx()
```
chokobole, line 160: Actually, throughout the codes, we use like below.

```suggestion
	ctx, cancel := context.WithTimeout(context.Background(), c.cfg.NetworkTimeout)
	defer cancel()
```

components/validator/challenger.go
```diff
@@ -65,191 +93,413 @@ func NewChallenger(ctx context.Context, cfg Config, l log.Logger) (*Challenger,
...
+	status, err := c.cfg.RollupClient.SyncStatus(ctx)
```
chokobole, line 161: You don't check the error here.

components/validator/challenger.go
```diff
@@ -65,191 +93,413 @@ func NewChallenger(ctx context.Context, cfg Config, l log.Logger) (*Challenger,
...
+
+	toBlock := new(big.Int).SetUint64(status.CurrentL1.Number)
+	// fromBlock = max(1, startBlockOfFinalizationWindow)
```
chokobole, line 164: Please remove the comment.

components/validator/challenger.go
```diff
@@ -65,191 +93,413 @@ func NewChallenger(ctx context.Context, cfg Config, l log.Logger) (*Challenger,
...
+	fromBlock := math.BigMax(new(big.Int).Sub(toBlock, new(big.Int).Div(c.finalizationPeriodSeconds, big.NewInt(12))), big.NewInt(1))
```
chokobole, line 165: Please use `L1BlockTime` variable which is read from confing.

components/validator/challenger.go
```diff
@@ -65,191 +93,413 @@ func NewChallenger(ctx context.Context, cfg Config, l log.Logger) (*Challenger,
...
-			c.checkpoint = common.Big1
-		} else if waitingOutputNum.Cmp(common.Big0) == 0 {
-			c.checkpoint = latestOutputIndex
-			return nil, nil
-		} else {
-			c.checkpoint = new(big.Int).Sub(nextOutputIndex, waitingOutputNum)
-		}
+	logs, err := c.l1Client.FilterLogs(context.Background(), query)
+	if err != nil {
+		c.log.Error("failed to get event logs related to outputs", "err", err)
```
chokobole, line 181: In this case, this should repeat the process after 1 minutes or somethig...?

components/validator/challenger.go
```diff
@@ -65,191 +93,413 @@ func NewChallenger(ctx context.Context, cfg Config, l log.Logger) (*Challenger,
...
+		// for OutputSubmitted event
+		if vLog.Address == c.cfg.L2OutputOracleAddr {
+			outputIndex := new(big.Int).SetBytes(vLog.Topics[2][:])
 
-		knownOutput, err := c.OutputAtBlockSafe(output.L2BlockNumber.Uint64(), false)
-		if err != nil {
-			return nil, err
+			outputRange, err := c.ValidateOutput(outputIndex)
+			if err != nil {
+				c.log.Warn("unable to validate output", "err", err, "outputIndex", outputIndex)
```
chokobole, line 192: This error happens only due to unresponsive L2 Geth or L2 nodes. So it can be helpful to record the output or relevant information when such errors occur. Then I think we should retry to call `ValidateOutput` once again.

components/validator/challenger.go
```diff
@@ -65,191 +93,413 @@ func NewChallenger(ctx context.Context, cfg Config, l log.Logger) (*Challenger,
...
-		}
+	for _, vLog := range logs {
+		// for OutputSubmitted event
+		if vLog.Address == c.cfg.L2OutputOracleAddr {
+			outputIndex := new(big.Int).SetBytes(vLog.Topics[2][:])
 
-		knownOutput, err := c.OutputAtBlockSafe(output.L2BlockNumber.Uint64(), false)
-		if err != nil {
-			return nil, err
+			outputRange, err := c.ValidateOutput(outputIndex)
```
chokobole, line 190: For optimization, can we load all the local and remote outputs as a batch? This requires so many IO round.

components/validator/challenger.go
```diff
@@ -65,191 +93,413 @@ func NewChallenger(ctx context.Context, cfg Config, l log.Logger) (*Challenger,
...
+				continue
+			}
+			if outputRange == nil {
+				continue
+			}
+
+			// check if the challenge is in progress already.
+			isInProgress, err := c.IsChallengeInProgress(outputIndex)
+			if err != nil {
+				c.log.Warn("unable to get the status of challenge", "err", err, "outputIndex", outputIndex)
```
chokobole, line 202: Ditto, we need to record this and retry

components/validator/challenger.go
```diff
@@ -65,191 +93,413 @@ func NewChallenger(ctx context.Context, cfg Config, l log.Logger) (*Challenger,
...
+				continue
+			}
+			if isInProgress {
+				c.log.Warn("found invalid output, but is already in progress", "outputIndex", outputIndex)
```
chokobole, line 206: is it warning? I think this should be info.

components/validator/challenger.go
```diff
@@ -65,191 +93,413 @@ func NewChallenger(ctx context.Context, cfg Config, l log.Logger) (*Challenger,
...
+				continue
+			}
+
+			// if there is no challenge on invalid output, create a new challenge
+			tx, err := c.CreateChallenge(outputRange)
+			if err != nil {
+				c.log.Error("failed to create createChallenge tx", "err", err)
```
chokobole, line 213: ditto, we need to do once again.

components/validator/challenger.go
chokobole: Plus, the scope of cancel context is too long. Just imagine below for-loop loops over 7 days output submission!

components/validator/challenger.go
```diff
@@ -65,191 +93,413 @@ func NewChallenger(ctx context.Context, cfg Config, l log.Logger) (*Challenger,
...
+		return
+	}
+
+	isAsserter := challenge.Asserter == c.cfg.TxManager.From()
+	isChallenger := challenge.Challenger == c.cfg.TxManager.From()
+
+	// not related challenge
+	if !isAsserter && !isChallenger {
+		c.log.Error("not related challenge", "outputIndex", outputIndex)
+		return
```
chokobole, line 359: Can't we check this validation earlier..?

components/validator/challenger.go
```diff
@@ -65,191 +93,413 @@ func NewChallenger(ctx context.Context, cfg Config, l log.Logger) (*Challenger,
...
-				return c.Bisect()
-			case chal.StatusChallengerTimeout:
-				challengeId, err := c.GetChallengeInProgress()
-				if err != nil {
-					return nil, err
-				}
+	// get challenge
+	challenge, err := c.GetChallenge(outputIndex)
+	if err != nil {
+		c.log.Error("failed to get challenge", "err", err, "outputIndex", outputIndex)
```
chokobole, line 349: This should call `handleChallenge` itself once again.

components/validator/challenger.go
```diff
@@ -65,191 +93,413 @@ func NewChallenger(ctx context.Context, cfg Config, l log.Logger) (*Challenger,
...
-		if !c.cfg.ChallengerDisabled {
-			switch status {
-			case chal.StatusChallengerTurn:
-				return c.Bisect()
-			case chal.StatusAsserterTimeout:
-				return c.AsserterTimeout()
-			case chal.StatusProveReady:
-				return c.ProveFault()
+			// if the status is NONE or CHALLENGER_TIMEOUT
+			if status == chal.StatusNone || status == chal.StatusChallengerTimeout {
```
chokobole, line 377: I think it's better to create `isInProgress()` function for go version!

components/validator/challenger.go
```diff
@@ -65,191 +93,413 @@ func NewChallenger(ctx context.Context, cfg Config, l log.Logger) (*Challenger,
...
+	// not related challenge
+	if !isAsserter && !isChallenger {
+		c.log.Error("not related challenge", "outputIndex", outputIndex)
+		return
+	}
 
-				// TODO(pangssu): Is it necessary to submit challengerTimeout transaction?
-				c.log.Info("challenger timed out", "challengeId", challengeId)
-				return nil, nil
+	ticker := time.NewTicker(c.cfg.PollInterval)
```
chokobole, line 362: I think you should separate this PollInterval from the PollInterval that is used by a validator. This is because output root is submitted almost every 1 hour and you should poll rarely. But for the challenge, you should poll quite a lot because it should react to the opponent.

components/validator/challenger.go
```diff
@@ -65,191 +93,413 @@ func NewChallenger(ctx context.Context, cfg Config, l log.Logger) (*Challenger,
...
+				case chal.StatusChallengerTurn:
+					tx, err := c.Bisect(outputIndex)
+					if err != nil {
+						c.log.Error("challenger: failed to create bisect tx", "err", err, "outputIndex", outputIndex)
+						break ChallengeLoop
+					}
+					c.submitChallengeTx(tx)
+				case chal.StatusAsserterTimeout, chal.StatusProveReady:
+					tx, err := c.ProveFault(outputIndex)
+					if err != nil {
```
chokobole, line 412: Please just know that this part may take so long.

components/validator/validator.go
```diff
@@ -86,31 +88,39 @@ func NewValidator(parentCtx context.Context, cfg Config, l log.Logger, m metrics
...
+		txCandidatesChan: txCandidatesChan,
+		metr:             m,
 	}, nil
 }
 
 func (v *Validator) Start() {
 	v.l.Info("starting Validator")
+
+	if err := v.challenger.Start(); err != nil {
+		return
```
chokobole, line 121: Please add a log here or make `Start()` return error

e2e/actions/l2_challenger.go
```diff
@@ -107,18 +107,17 @@ func (c *L2Challenger) ActBisect(t Testing) common.Hash {
...
 
-	if status == chal.StatusAsserterTimeout {
-		tx, err = c.challenger.AsserterTimeout()
-	} else if status == chal.StatusChallengerTimeout {
-		challengeId, err := c.challenger.LatestChallengeId()
-		require.NoError(t, err)
-		tx, err = c.challenger.ChallengerTimeout(challengeId)
+	//if status == chal.StatusAsserterTimeout {
+	//	tx, err = c.challenger.AsserterTimeout()
+	//} else
```
chokobole, line 118: Please remove the comments.

components/validator/challenger.go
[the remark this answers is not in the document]
0xHansLee: As @chokobole mentioned offline, I will remove `done` channel and use context instead.

components/validator/challenger.go
0xHansLee: This was the comment for `fromBlock` how to calculate it. I modified to descriptive way

components/validator/challenger.go
0xHansLee: Good. I'll optimize the validation of multiple outputs.

components/validator/challenger.go
```diff
@@ -24,28 +32,48 @@ type ProofFetcher interface {
...
+	ctx  context.Context
+	done chan struct{}
 
 	callOpts *bind.CallOpts
 	txOpts   *bind.TransactOpts
 
-	l2ooContract      *bindings.L2OutputOracle
+	l1Client *ethclient.Client
+
+	l2ooContract      *bindings.L2OutputOracleCaller
```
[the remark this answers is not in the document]
0xHansLee: I didn't know that the diff between `L2OutputOracle` and `L2OutputOracleCaller`. Just use caller as done in Optimism. I'll change it back to `L2OutputOracle`, not caller. Thx

components/validator/challenger.go
0xHansLee: The `L1BlockTime` is not in config.
When I add the `L1BlockTime` to rollup config, it makes many file changes.
Thus I mark `TODO` and will do in another PR.

components/validator/challenger.go
0xHansLee: An optimization of validation of multiple outputs may lead to some changes in kroma-node and L2 output oracle contract. Thus, I mark `TODO` and will handle it in another PR.

components/validator/challenger.go
0xHansLee: This check has done before handling challenge using the addresses of asserter and challenger from event.

0xHansLee:
Sorry for you to wait. I made some changes since this PR has been first opened.

Discussed about retry with @Pangssu, and we concluded that all the process for validator should succeed.
Thus, when errors occur in all the process related to validator, retry it until succeed.
I also added access list for submitting challenge related transactions.

There are still somethings to be optimized. Optimizations that can cause big changes were `TODO` marked for future work.
If you think that some optimizations should be done in this PR, please leave comments.

components/validator/challenger.go
```diff
@@ -24,232 +31,506 @@ type ProofFetcher interface {
...
+	for {
+	Loop:
+		select {
+		case <-c.ctx.Done():
+			return
+		default:
+			status, err := c.cfg.RollupClient.SyncStatus(c.ctx)
+			if err != nil {
+				c.log.Error("failed to get sync status", "err", err)
+				time.Sleep(time.Minute)
```
chokobole, line 170: Why do we need to sleep here?

components/validator/challenger.go
```diff
@@ -24,232 +31,506 @@ type ProofFetcher interface {
...
+		},
+	}
 
-	c.checkpoint = new(big.Int).Add(latestOutputIndex, common.Big1)
+	for {
+	Loop:
+		select {
+		case <-c.ctx.Done():
+			return
+		default:
```
chokobole, line 246: ditto

components/validator/challenger.go
```diff
@@ -24,232 +31,506 @@ type ProofFetcher interface {
...
-				"start", start,
-				"end", end,
-				"outputRoot", common.BytesToHash(output.OutputRoot[:]),
-			)
+				// for ChallengeCreated event
+				if vLog.Address == c.cfg.ColosseumAddr {
+					// handle challenge
+					outputIndex := new(big.Int).SetBytes(vLog.Topics[1][:])
+					asserter := common.BytesToAddress(vLog.Topics[2][:])
+					challenger := common.BytesToAddress(vLog.Topics[3][:])
```
chokobole, line 214: How about making struct for `ChallengeTopic` and share the codes with below.

```go
type ChallengeCreatedEvent struct {
  OutputIndex *big.Int
  Asserter common.Address
  Challenger common.Address
}

func NewChallengeCreatedEvent(log *types.Log) ChallengeCreatedEvent {
  // ...
}
```

components/validator/challenger.go
```diff
@@ -24,232 +31,506 @@ type ProofFetcher interface {
...
+				case vLog := <-logs:
+					expected := vLog.Topics[1]
+					outputIndex := new(big.Int).SetBytes(vLog.Topics[2][:])
+					l2BlockNumber := new(big.Int).SetBytes(vLog.Topics[3][:])
+					c.log.Info("validating output", "l2BlockNumber", l2BlockNumber, "outputRoot", expected.Hex(), "outputIndex", outputIndex)
+
+					// validate all outputs in between the checkpoint and the current outputIndex
+					for i := c.checkpoint; i.Cmp(outputIndex) != 1; new(big.Int).Add(i, common.Big1) {
+						c.wg.Add(1)
+						c.handleOutput(i)
```
chokobole, line 270:
```suggestion
						go c.handleOutput(i)
```

components/validator/challenger.go
0xHansLee: I changed to use `time.Ticker` 👍

chokobole (approved):
LGTM

Pangssu (approved):
LGTM

components/validator/challenger.go
```diff
@@ -24,232 +31,511 @@ type ProofFetcher interface {
...
-}
+func (c *Challenger) Start() error {
+	c.log.Info("start challenger")
 
-func (c *Challenger) OutputAtBlockSafe(blockNumber uint64, includeNextBlock bool) (*eth.OutputResponse, error) {
-	ctx, cancel := context.WithTimeout(c.ctx, c.cfg.NetworkTimeout)
-	defer cancel()
-	output, err := c.cfg.RollupClient.OutputAtBlock(ctx, blockNumber, includeNextBlock)
+	// if checkpoint is behind the latest output index, scan the previous outputs from the checkpoint
+	latestOutputIndex, err := c.l2ooContract.LatestOutputIndex(c.callOpts)
```
seolaoh, line 126: When I try to up devnet, I get an error here and fail to start the validator node, because this call reverts when there is no output submitted in `l2ooContract`.

0xHansLee: Thank you. I will cover the case that there is no output at genesis.

components/validator/challenger.go
```diff
@@ -24,232 +31,511 @@ type ProofFetcher interface {
...
+			}
+
+			query := ethereum.FilterQuery{
+				FromBlock: fromBlock,
+				ToBlock:   toBlock,
+				Addresses: addresses,
+				Topics:    [][]common.Hash{topics},
+			}
+
+			logs, err := c.l1Client.FilterLogs(context.Background(), query)
```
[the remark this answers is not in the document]
0xHansLee: Changed to use `c.ctx`. Thx!

components/validator/challenger.go
```diff
@@ -24,232 +31,515 @@ type ProofFetcher interface {
...
 
-func (c *Challenger) IsChallengeInProgress() (bool, error) {
-	return c.colosseumContract.IsInProgress(c.callOpts)
-}
+func (c *Challenger) Start() error {
+	c.log.Info("start challenger")
+
+	// if checkpoint is behind the latest output index, scan the previous outputs from the checkpoint
+	latestOutputIndex, err := c.l2ooContract.LatestOutputIndex(c.callOpts)
+	if err.Error() == "execution reverted" { // no output submitted
```
kangsorang, line 127: The handling based on error message wording seems ambiguous, and it would be nice to see some clarification on the reverted behavior.

0xHansLee: I agree. As you said, it's quite implicit so I changed to check if the `NextOutputIndex` is equal to zero or not.
(If the `NextOutputIndex` is equal to 0, there is no output submitted)

seolaoh (approved):
LGTM

components/validator/validator.go
```diff
@@ -113,34 +119,46 @@ func NewValidator(parentCtx context.Context, cfg Config, l log.Logger, m metrics
...
 		if err := v.l2os.Start(); err != nil {
-			v.l.Error("cannot start l2 output submitter: %w", err)
-			return
+			return fmt.Errorf("cannot start l2 output submitter: %w", err)
 		}
 	}
 
+	if err := v.challenger.Start(); err != nil {
+		return fmt.Errorf("cannot start challenger: %w", err)
+	}
```
seolaoh, line 133: Don't we need to check the following?
```suggestion
	if !v.cfg.OutputSubmitterDisabled || !v.cfg.ChallengerDisabled {
		if err := v.challenger.Stop(); err != nil {
			return fmt.Errorf("failed to stop challenger: %w", err)
		}
	}
```
Or check it in config check function in `/components/validator/config.go` L46?

seolaoh: Also in `Stop()`.

seolaoh, line 133: Don't we need to check the following?
```suggestion
	if !v.cfg.OutputSubmitterDisabled || !v.cfg.ChallengerDisabled {
		if err := v.challenger.Start(); err != nil {
			return fmt.Errorf("cannot start challenger: %w", err)
		}
	}
```
Or check it in config check function in `/components/validator/config.go` L46?
Also in `Stop()`.

seolaoh (changes requested):

components/validator/validator.go
0xHansLee: I got your point. 
I think validator should set `true` at least either `OutputSubmitterDisabled` or `ChallengerDisabled`. (That is, do check in config)
And the challenger should start whichever the mode is. wdyt?

seolaoh: I do agree! Then it would be better to check if at least one of the flags is on in config.

kangsorang:
On a slightly different topic, is there a reason why the flags for the behavior are set to disabled instead of enabled? I think it's a little confusing to have a ! in front of the disabled flag every time.

components/validator/validator.go
0xHansLee: Added in [0cddd32](https://github.com/kroma-network/kroma/pull/59/commits/0cddd322255e32efd3865e0684a22515b1d8c1c5)

components/validator/challenger.go
```diff
@@ -24,232 +31,518 @@ type ProofFetcher interface {
...
+
+	// if challenge mode on, subscribe L2 output submission events
+	if !c.cfg.ChallengerDisabled {
+		c.wg.Add(1)
+		go c.subscribeL2OutputSubmitted()
+	}
+
+	// subscribe challenge creation events
+	c.wg.Add(1)
+	go c.subscribeChallengeCreated()
```
0xHansLee, line 149: `scanPrevOutputs()` and `subscribeChallengerCreated()` seem that those should be the function of the `Validator`, not `Challenger`.
Because those two should be called regardless of whether the challenger is enabled or not.

0xHansLee: I leave it as is for now, and will refactor for future.

0xHansLee:
> On a slightly different topic, is there a reason why the flags for the behavior are set to disabled instead of enabled? I think it's a little confusing to have a ! in front of the disabled flag every time.

I'm not sure because I wasn't involved at that time, but I think it was because the validator is designed to be both output submitter and challenger by default.

seolaoh (approved):

components/validator/challenger.go
```diff
@@ -24,232 +31,493 @@ type ProofFetcher interface {
...
+					// validate all outputs in between the checkpoint and the current outputIndex
+					for i := c.checkpoint; i.Cmp(ev.L2OutputIndex) != 1; new(big.Int).Add(i, common.Big1) {
+						c.wg.Add(1)
+						go c.handleOutput(i)
+					}
+					c.checkpoint = ev.L2OutputIndex
+				case <-c.ctx.Done():
+					return
+				}
+			}
```
0xHansLee, line 273: @kangsorang 
As you suggested, I made this simpler by using the subscribe function which is generated by bindings.
With this, we don't need to handle error from subscriber any more :)

components/validator/challenger.go
```diff
@@ -24,232 +31,493 @@ type ProofFetcher interface {
...
+	ticker := time.NewTicker(time.Minute)
+	defer ticker.Stop()
+
+	for {
+	Loop:
+		select {
+		case <-ticker.C:
+			opts := &bind.WatchOpts{Context: c.ctx}
+			eventChan := make(chan *bindings.L2OutputOracleOutputSubmitted)
+			_, err := c.l2ooContract.WatchOutputSubmitted(opts, eventChan, nil, nil, nil)
```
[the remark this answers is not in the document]
0xHansLee: You're right. I thought that `Watch*` functions do everything including unsubscribe() but it wasn't.
I fixed it please review again. Thanks in advance!

kangsorang:
LGTM

components/validator/challenger.go
```diff
@@ -24,232 +32,490 @@ type ProofFetcher interface {
...
+}
 
-				// TODO(pangssu): Is it necessary to submit challengerTimeout transaction?
-				c.log.Info("challenger timed out", "challengeId", challengeId)
-				return nil, nil
+// subscribeL2OutputSubmitted subscribes the OutputSubmitted event from L2OutputOracle contract.
+// If the L2 output root is invalid, create challenge.
+// This function should be called only when challenger mode is on.
+func (c *Challenger) subscribeL2OutputSubmitted() {
+	c.wg.Done()
```
kangsorang, line 275: I think this part should be called by the defer routine.

components/validator/challenger.go
```diff
@@ -24,232 +32,490 @@ type ProofFetcher interface {
...
-			switch status {
-			case chal.StatusChallengerTurn:
-				return c.Bisect()
-			case chal.StatusAsserterTimeout:
-				return c.AsserterTimeout()
-			case chal.StatusProveReady:
-				return c.ProveFault()
+// subscribeChallengeCreated subscribes the ChallengeCreated event from Colosseum contract and handle challenge.
+func (c *Challenger) subscribeChallengeCreated() {
+	c.wg.Done()
```
kangsorang, line 295: Ditto

components/validator/challenger.go
0xHansLee: Oh, mistake. Thanks!

kangsorang:
LGTM

kangsorang (approved):
LGTM

Issue #19, opened 2023-04-05
Allow multiple validators to submit output


chokobole:
### Issue Type

Feature Request

### Kroma Version

0.1.0

### OS Platform and Distribution

Linux Ubuntu 22.04

### Go version

go1.19.3 linux/amd64

### Node version

v16.14.2

### Changes

Currently a single validator only submits output, which is called _validation_. For the next release, we want to contain permissionless validation. 

**TL;DR** 

1. `Fee vaults` would be divided into three categories based on its purpose. `Proposer Reward Vault`, `Protocol Vault` and `Validator Reward Vault`.
2. Validators need to submit output with a _bond_ of at least minimum bond amount.
3. Validators receive a reward in exchange for submitting an output.
4. Validators needs to initiate a challenge with a _bond_.
5. The amount of minimum bond and reward can dynamically vary depending on the protocol.

Optimism distinguishes between three different types of funding sources: `ProposerFeeVault`, `BaseFeeVault`, and `RollupFeeVault`, based on their origin. In the upcoming release of Kroma, funds will be divided into `ProposerRewardVault`, `ProtocolVault`, and `ValidatorRewardVault`, based on their purpose. The `ProposerRewardVault` will be similar to the existing `RollupFeeVault`. Of the total fee (sum of base fee and priority fee), a proportion _R_ (`Validator Reward Ratio`) will go to the `ValidatorRewardVault`, while the remaining amount will go to the `ProtocolVault`. The funds in the `ValidatorRewardVault` will be used solely for rewarding validators, while the `ProtocolVault` will be used for the development of the protocol.

_R_ is designed considering the following:

- _R_ is not fixed but dynamically set depending on the situation.
- _R_ is updated in the L1Block contract via DepositTx when every L2 block is created.
- However, we haven't found any factors that would change the current _R_, so we'll need to do some research and temporarily fix it at a certain value. (probably 0.2)

To participate in the validation process, a validator must stake a certain amount of bond, but if the bonding amount is too high, the profitability will decrease, and if it is too low, it may lead to abuse. Therefore, to ensure that the staking profitability is similar to that of ETH staking, it is important to set $R_{min}$ and $R_{max}$ appropriately. If the output submitted by a validator is successfully finalized, they can claim a share of the amount present in the `ValidatorRewardVault` at that time proportional to their contribution. However, there is a submission timeout for validation, and if the output is not submitted within that timeout, the trusted validator will receive the reward.

[...]

PR #63, opened 2023-06-09
feat: add validator deposit, withdraw, unbond commands

seolaoh:
# Description

Added balance managing commands to `validator` app to provide the ability to deposit, withdraw, and unbond to `ValidatorPool` from CLI. Also added validator deposit check to avoid performing unnecessary processes if the validator does not have enough deposits when trying to submit output.
I added validator deposit guide doc into `specs/meta`, if you have a better idea for where to place this file, please let me know.

When providing `accessList` that is not actually accessed, the transaction may not be executed due to exceeding the estimated gas limit. I fixed incorrect storage key of output submit tx, and added transaction receipt fail status handling in `txmgr`.

Additionally, slightly changed `docker-compose.yml` to fix docker image and environment variables correctly.

In devnet, I verified that the added commands work fine.

Related #19


components/validator/l2_output_submitter.go
```diff
@@ -298,12 +338,11 @@ func submitL2OutputTxData(abi *abi.ABI, output *eth.OutputResponse, bondAmount u
...
+	// TODO(sara): add correct accessList using `eth_createAccessList` RPC method.
+	// When you provide incorrect accessList, the transaction might be failed.
 	accessList := types.AccessList{
 		types.AccessTuple{
 			Address: l.cfg.ValidatorPoolAddr,
-			StorageKeys: []common.Hash{
-				common.HexToHash("0000000000000000000000000000000000000000000000000000000000000036"),
-			},
 		},
 	}
```
seolaoh, line 347: I removed the storage key of access list. Because due to the existing access list, the transaction is confirmed but the status is confirmed as failed. This is probably because the storage key is incorrect, or all the required keys are not provided. We might to look up geth code later, but I don't think this is a critical issue now, so I've marked it TODO.

components/validator/l2_output_submitter.go
```diff
@@ -168,11 +179,21 @@ func (l *L2OutputSubmitter) trySubmitL2Output() error {
...
 // canSubmit checks if submission interval has elapsed and selected for next validator.
-func (l *L2OutputSubmitter) canSubmit(ctx context.Context) (*big.Int, bool, error) {
-	currentBlockNumber, nextBlockNumber, err := l.fetchBlockNumbers(ctx)
+func (l *L2OutputSubmitter) canSubmit() (*big.Int, bool, error) {
+	hasEnoughDeposit, err := l.checkDeposit()
 	if err != nil {
 		return nil, false, err
 	}
+	if !hasEnoughDeposit {
+		l.retryAfter(1)
```
chokobole, line 188: What if we use `time.Duration` as the argument to indicate whether it's in seconds or not? It would make it easier to understand when reading the code or following the function calls.

```go
func (l *L2OutputSubmitter) retryAfter(d time.Duration) {
	time.AfterFunc(d, func() {
		l.submitChan <- struct{}{}
	})
}
```

specs/meta/validator-deposit.md
```diff
@@ -0,0 +1,78 @@
...
+> --l1-eth-rpc <l1-eth-rpc> \
+> --mnemonic <mnemonic> \
+> --hd-path <hd-path> \
+> --rollup-rpc "" \ # empty required flags
+> --l2oo-address "" \
+> --colosseum-address "" \
+> --challenger.poll-interval 0s \
+> balance withdraw \
+> --withdraw-amount <withdraw-amount-wei> # must be set
+```
```
chokobole, line 59:
```suggestion
```shell
> go run ./components/validator/cmd/main.go \
  --valpool-address <validator-pool-address> \ # must be set
  --l1-eth-rpc <l1-eth-rpc> \
  --mnemonic <mnemonic> \
  --hd-path <hd-path> \
  --rollup-rpc "" \ # empty required flags
  --l2oo-address "" \
  --colosseum-address "" \
  --challenger.poll-interval 0s \
  balance withdraw \
  --withdraw-amount <withdraw-amount-wei> # must be set
```
```

File: specs/meta/validator-deposit.md [hunk 4]
```diff
@@ -0,0 +1,78 @@
...
+> --l1-eth-rpc <l1-eth-rpc> \
+> --mnemonic <mnemonic> \
+> --hd-path <hd-path> \
+> --rollup-rpc "" \ # empty required flags
+> --l2oo-address "" \
+> --colosseum-address "" \
+> --challenger.poll-interval 0s \
+> balance deposit \
+> --deposit-amount <deposit-amount-wei> # must be set
+```
```
chokobole, line 39:
```suggestion
```shell
> go run ./components/validator/cmd/main.go \
  --valpool-address <validator-pool-address> \ # must be set
  --l1-eth-rpc <l1-eth-rpc> \
  --mnemonic <mnemonic> \
  --hd-path <hd-path> \
  --rollup-rpc "" \ # empty required flags
  --l2oo-address "" \
  --colosseum-address "" \
  --challenger.poll-interval 0s \
  balance deposit \
  --deposit-amount <deposit-amount-wei> # must be set
```
```

File: specs/meta/validator-deposit.md [hunk 5]
```diff
@@ -0,0 +1,78 @@
...
+> --valpool-address <validator-pool-address> \ # must be set
+> --l1-eth-rpc <l1-eth-rpc> \
+> --mnemonic <mnemonic> \
+> --hd-path <hd-path> \
+> --rollup-rpc "" \ # empty required flags
+> --l2oo-address "" \
+> --colosseum-address "" \
+> --challenger.poll-interval 0s \
+> balance unbond
+```
```
chokobole, line 78:
```suggestion
```shell
> go run ./components/validator/cmd/main.go \
  --valpool-address <validator-pool-address> \ # must be set
  --l1-eth-rpc <l1-eth-rpc> \
  --mnemonic <mnemonic> \
  --hd-path <hd-path> \
  --rollup-rpc "" \ # empty required flags
  --l2oo-address "" \
  --colosseum-address "" \
  --challenger.poll-interval 0s \
  balance unbond
```
```

File: ops-devnet/devnet-up.sh [hunk 6]
```diff
@@ -97,6 +98,25 @@ fi
...
+  echo "Deposit into ValidatorPool to be a validator..."
+  go run cmd/main.go \
+    --l1-eth-rpc $L1_URL \
+    --rollup-rpc "" \
+    --l2oo-address "" \
+    --colosseum-address "" \
+    --valpool-address $VALPOOL_ADDRESS \
+    --challenger.poll-interval 0s \
+    --mnemonic "$VALIDATOR_MNEMONIC" \
+    --hd-path $VALIDATOR_HD_PATH \
```
chokobole, line 116: Is there a way to exclude unnecessary flags when running balance subcommand?

components/validator/cmd/balance/cmd.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/kroma-network/kroma/utils/service/txmgr/metrics"
+)
+
+var Subcommands = cli.Commands{
+	{
+		Name:  "deposit",
+		Usage: "Deposit ETH into ValidatorPool to be used as bond",
+		Flags: []cli.Flag{
+			cli.Uint64Flag{
+				Name:     "deposit-amount",
```
Pangssu, line 24: simply, `amount`

components/validator/cmd/balance/cmd.go
```diff
@@ -0,0 +1,133 @@
...
+			},
+		},
+		Action: deposit,
+	},
+	{
+		Name:  "withdraw",
+		Usage: "Withdraw ETH from ValidatorPool",
+		Flags: []cli.Flag{
+			cli.Uint64Flag{
+				Name:     "withdraw-amount",
```
Pangssu, line 36: simply, `amount`

components/validator/cmd/main.go
```diff
@@ -28,6 +29,13 @@ func main() {
 	app.Description = "Service for generating and submitting L2 output checkpoints to the L2OutputOracle contract as an L2 Output Submitter, " + "detecting and correcting invalid L2 outputs as a Challenger to ensure the integrity of the L2 state."
 
 	app.Action = curryMain(Version)
+	app.Commands = []cli.Command{
+		{
+			Name:        "balance",
+			Subcommands: balance.Subcommands,
+		},
+	}
```
Pangssu, line 37: I think it looks better that `deposit`, `withdraw` and `unbond` subcommands are declared here.

components/validator/l2_output_submitter.go
```diff
@@ -78,13 +79,22 @@ func NewL2OutputSubmitter(parentCtx context.Context, cfg Config, l log.Logger, m
...
+	cCancel()
+
+	cCtx, cCancel = context.WithTimeout(ctx, cfg.NetworkTimeout)
+	defer cCancel()
+	callOpts = utils.NewSimpleCallOpts(cCtx)
+	minBondAmount, err := valpoolContract.MINBONDAMOUNT(callOpts)
+	if err != nil {
+		cancel()
+		return nil, fmt.Errorf("failed to get min bond amount: %w", err)
+	}
```
Pangssu, line 97: I don't think the `minBondAmount` is necessary, because the `bondAmount` is declared in config, and it will try to bond with that amount..

components/validator/l2_output_submitter.go
```diff
@@ -202,8 +223,27 @@ func (l *L2OutputSubmitter) canSubmit(ctx context.Context) (*big.Int, bool, erro
...
+	cCtx, cCancel := context.WithTimeout(l.ctx, l.cfg.NetworkTimeout)
+	defer cCancel()
+	from := l.cfg.TxManager.From()
+	callOpts := utils.NewCallOptsWithSender(cCtx, from)
+	balance, err := l.valpoolContract.BalanceOf(callOpts, from)
+	if err != nil {
+		return false, fmt.Errorf("failed to fetch validator deposit amount: %w", err)
+	}
+
+	if balance.Cmp(l.minBondAmount) == -1 {
```
Pangssu, line 236: The amount actually needed is the bondAmount in the config, not the `minBondAmount`.

ops-devnet/devnet-up.sh
```diff
@@ -105,6 +125,8 @@ VALPOOL_ADDRESS="0x6900000000000000000000000000000000000005"
   L2OO_ADDRESS="$L2OO_ADDRESS" \
     COLOSSEUM_ADDRESS="$COLOSSEUM_ADDRESS" \
     VALPOOL_ADDRESS="$VALPOOL_ADDRESS" \
+    VALIDATOR_MNEMONIC="$VALIDATOR_MNEMONIC" \
+    VALIDATOR_HD_PATH="$VALIDATOR_HD_PATH" \
```
Pangssu, line 129: Is there any way to avoid setting these values as environment variables?

ops-devnet/docker-compose.yml
```diff
@@ -30,7 +30,7 @@ services:
 
   l2:
     pid: host # allow debugging
-    image: kromanetwork/geth:dev
+    image: kromanetwork/geth
```
Pangssu, line 33: Please modify this with a tag for the build that corresponds to the version of go-ethereum referenced in `go.mod`.

components/validator/l2_output_submitter.go
chokobole: This is strange. Could you share me with how to test this?

specs/meta/validator-deposit.md
```diff
@@ -0,0 +1,78 @@
...
+- [Withdraw from `ValidatorPool`](#withdraw-from-validatorpool)
+- [Try unbond in `ValidatorPool`](#try-unbond-in-validatorpool)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+We want the [validator](../../components/validator/) role to be decentralized. Like how the PoS mechanism works, to
+achieve this, the validator needs to bond ETH at every
+[output submission](../validations.md#submitting-l2-output-commitments). A Validator can deposit into the contract
+`ValidatorPool` at once for
+convenience. This guide teaches you how to deposit, withdraw, and try to unbond in `ValidatorPool`.
```
0xHansLee, line 19: The `deposit` is not for convenience. It is required to bond ETH for output submission.
I think it would be better to describe two steps of deposit and bond specifically.
Also, we need to mention that the bonded ETH will be unbonded after the finalization of the output.

0xHansLee (approved):

components/validator/l2_output_submitter.go
```diff
@@ -202,8 +223,27 @@ func (l *L2OutputSubmitter) canSubmit(ctx context.Context) (*big.Int, bool, erro
...
+		l.log.Info("validator deposit is less than minimum bond amount", "minBondAmount", l.minBondAmount, "deposit", balance)
```
0xHansLee, line 237: maybe warn or error log

ops-devnet/devnet-up.sh
seolaoh: I cannot find any feature provided by `urfave/cli` to override or omit required flags from a parent command in a subcommand, so I guess the only way to omit unnecessary flags in a subcommand would be to replace the required flag with an optional one, but I don't think that's a good idea.

components/validator/l2_output_submitter.go
seolaoh: When providing `accessList` that is not actually accessed, the transaction may not be executed due to exceeding the estimated gas limit. I fixed incorrect storage key of output submit tx, and added transaction receipt fail status handling in `txmgr`.

specs/meta/validator-deposit.md
seolaoh: I've added a brief explanation as you said. Since this is a guide to utilizing the commands, a more detailed explanation should probably be included in `validations.md`.

ops-devnet/devnet-up.sh
seolaoh: Maybe via file input? Do you have any ideas?

components/validator/cmd/main.go
```diff
@@ -28,6 +29,43 @@ func main() {
 	app.Description = "Service for generating and submitting L2 output checkpoints to the L2OutputOracle contract as an L2 Output Submitter, " + "detecting and correcting invalid L2 outputs as a Challenger to ensure the integrity of the L2 state."
 
 	app.Action = curryMain(Version)
+	app.Commands = []cli.Command{
+		{
+			Name: "balance",
```
Pangssu, line 34: How about removing the `balance` command and having `deposit`, `withdraw`, and `unbond` come here?

Pangssu: like this.

`kroma-validator deposit <amount>`
`kroma-validator withdraw <amount>`

specs/meta/validator-deposit.md
0xHansLee: Oh, right. The detailed description should be in `validations.md` or `validators.md`.

components/validator/cmd/main.go
seolaoh: I made the changes you mentioned, but I didn't move all the functions in `balance/cmd.go` file to `validator/main.go` file to consider the possibility of other categories of commands being added in the future.

Pangssu: Good. This is what i expected.

ops-devnet/devnet-up.sh
0xHansLee: I agree that it'd be better to change not to use env var for mnemonic, but for that it seems that we need to change the way of config. The mnemonic (or private key) of batcher is also passed via env var.

seolaoh: Removed these lines by changing to run the `deposit` command after bringing up the validator node.

Pangssu (approved):
LGTM

PR #71, opened 2023-06-19
feat(validator): apply submission rounds to l2 output submitter

seolaoh:
# Description

Since submission interval has been divided into public round and priority round, apply it to l2 output submitter. Output submitter waits for public round to be reached when not selected for priority validator.
Additionally, retry interval and round buffer have been added to validator CLI flags.

Related https://github.com/kroma-network/kroma/issues/19


components/validator/l2_output_submitter.go
```diff
@@ -165,19 +167,19 @@ func (l *L2OutputSubmitter) trySubmitL2Output() error {
 		return fmt.Errorf("failed to submit l2 output transaction: %w", err)
 	}
 	l.metr.RecordL2OutputSubmitted(output.BlockRef)
-	l.retryAfter(1 * time.Second)
+	l.retryAfter(l.cfg.OutputSubmitterRetryInterval)
```
kangsorang, line 170: If a retryAfter call is needed even on a successful submit, it would seem to make sense to call retryAfter unconditionally in the trySubmitL2Output call.

components/validator/l2_output_submitter.go
```diff
@@ -268,31 +284,59 @@ func (l *L2OutputSubmitter) fetchBlockNumbers() (*big.Int, *big.Int, error) {
...
+
+type roundInfo struct {
+	isPublicRound       bool
+	isPriorityValidator bool
 }
 
-func (l *L2OutputSubmitter) isNextValidator() (bool, error) {
+// fetchCurrentRound fetches next validator address from ValidatorPool contract.
+// It returns if current round is public round, and if selected for priority validator if it's priority round.
+func (l *L2OutputSubmitter) fetchCurrentRound() (*roundInfo, error) {
```
chokobole, line 306: Since two booleans can fit within a 64-bit register, please simply return by copying.

```suggestion
func (l *L2OutputSubmitter) fetchCurrentRound() (roundInfo, error) {
```

components/validator/l2_output_submitter.go
seolaoh: Thank you for your suggestion. I will mark it as TODO because the changes might reflect E2E tests currently fixing.
I will change `trySubmitL2Output` function to return wait duration explicitly, and handle `retryAfter` function calls at once using the return value.

chokobole (approved):
LGTM

0xHansLee (approved):
lgtm

kangsorang (approved):
LGTM

PR #74, opened 2023-06-21
fix(validator): modify output to replace with when proving fault

seolaoh:
# Description
Challenger needs to submit valid output to replace invalid output with when proving fault. Since the challenger have submitted destination output root before, fixed it.

Related #19


chokobole (approved):
LGTM

0xHansLee (approved):
lgtm

Pangssu (approved):
LGTM

PR #79, opened 2023-06-22
feat(contracts): remove validation of previous output when challenge creation

seolaoh:
# Description

Removed to check if the first segment is matched with previous output root when creating challenge. This is because a delay attack is possible that takes advantage of the fact that only one challenge is possible on one output. Also added test of bisection with invalid segments.

Since the finalization period seconds of easel deploy config was too short, increased it from 60 seconds to 600 seconds.

Related #19


packages/contracts/contracts/test/Colosseum.t.sol
```diff
@@ -365,6 +360,42 @@ contract ColosseumTest is Colosseum_Initializer {
...
+        bytes32[] memory segments = _newSegments(
+            challenge.asserter,
+            challenge.turn + 1,
+            segStart,
+            segSize
+        );
+
+        vm.startPrank(challenge.asserter);
+
+        // invalid output of first segment
```
chokobole, line 384:
```suggestion
        // invalid output of the first segment
```

kangsorang (approved):
LGTM

0xHansLee (approved):
lgtm

chokobole (approved):
LGTM

PR #82, opened 2023-06-22
test(e2e): add challenge test

kangsorang:
# Description

- Add challenge e2e test

Test the challenge validation process with malicious validators, good challengers, and guardians.


components/validator/guardian.go
```diff
@@ -102,7 +102,7 @@ func (g *Guardian) handleValidationRequested() {
 }
 
 func (g *Guardian) processOutputValidation(event *bindings.SecurityCouncilValidationRequested) {
-	ticker := time.NewTicker(time.Minute)
+	ticker := time.NewTicker(10 * time.Second)
```
chokobole, line 105: Why did you change?

e2e/setup.go
```diff
@@ -586,21 +592,22 @@ func (cfg SystemConfig) Start(_opts ...SystemConfigOption) (*System, error) {
...
 		L1EthRpc:                     sys.Nodes["l1"].WSEndpoint(),
 		RollupRpc:                    sys.RollupNodes["proposer"].HTTPEndpoint(),
 		L2OOAddress:                  predeploys.DevL2OutputOracleAddr.String(),
 		ColosseumAddress:             predeploys.DevColosseumAddr.String(),
 		ValPoolAddress:               predeploys.DevValidatorPoolAddr.String(),
-		SecurityCouncilAddress:       predeploys.DevSecurityCouncilAddr.String(),
-		ChallengerPollInterval:       50 * time.Millisecond,
-		OutputSubmitterRetryInterval: 50 * time.Millisecond,
-		OutputSubmitterRoundBuffer:   30,
+		ChallengerPollInterval:       500 * time.Millisecond,
```
chokobole, line 602: Why did you change?

e2e/system_test.go
```diff
@@ -1769,6 +1772,79 @@ func TestStopStartBatcher(t *testing.T) {
...
+
+	sys, err := cfg.Start()
+	require.NoError(t, err, "Error starting up system")
+	defer sys.Close()
+
+	l1Client := sys.Clients["l1"]
+
+	// deposit to ValidatorPool to be a validator / challenger
+	err = cfg.DepositValidatorPool(l1Client, cfg.Secrets.TrustedValidator, big.NewInt(1_000_000_000))
+	require.NoError(t, err, "Error deposit to ValidatorPool")
```
chokobole, line 1791:
```suggestion
	require.NoError(t, err, "Error trusted validator deposit to ValidatorPool")
```

e2e/system_test.go
```diff
@@ -1769,6 +1772,79 @@ func TestStopStartBatcher(t *testing.T) {
...
+	err = cfg.DepositValidatorPool(l1Client, cfg.Secrets.Challenger, big.NewInt(1_000_000_000))
+	require.NoError(t, err, "Error deposit to ValidatorPool")
```
chokobole, line 1793:
```suggestion
	require.NoError(t, err, "Error challenger deposit to ValidatorPool")
```

e2e/setup.go
```diff
@@ -628,6 +646,53 @@ func (cfg SystemConfig) Start(_opts ...SystemConfigOption) (*System, error) {
...
+		return nil, fmt.Errorf("unable to init challenger config: %w", err)
+	}
+
+	// Replace to mock RPC client
+	cl, err = rpc.DialHTTP(challengerCliCfg.RollupRpc)
+	if err != nil {
+		return nil, fmt.Errorf("unable to init challenger rollup rpc client: %w", err)
+	}
+	rpcCl = client.NewBaseRPCClient(cl)
+	challengerMaliciousL2RPC := e2eutils.NewChallengerL2RPC(rpcCl)
```
0xHansLee, line 679: Challenger might be honest one, right?
```suggestion
	challengerL2RPC := e2eutils.NewChallengerL2RPC(rpcCl)
```

kangsorang: Yes, you're right, we need an honest name.

components/validator/guardian.go
kangsorang: I thought it could be a little shorter since it's a duration value that applies when an output validation request comes in, not all the time.

e2e/setup.go
kangsorang:  I don't see a need for settings with too short a cycle.
Since we don't inject all the timing variables used in the code at test time, lowering this value by a lot will not make the test time any shorter. Rather, it will cause a console load when debugging.

e2e/actions/l2_challenger_test.go
```diff
@@ -165,6 +174,10 @@ interaction:
 		case chal.StatusAsserterTimeout, chal.StatusReadyToProve:
 			txHash = challenger.ActProveFault(t, outputIndex)
 			includeL1Block(t, miner, challenger.address)
+		case chal.StatusProven:
+			// validate l2 out is needed?
+			guardian.ActConfirmTransaction(t, big.NewInt(0))
```
seolaoh, line 179:
```suggestion
			txHash = guardian.ActConfirmTransaction(t, big.NewInt(0))
```

e2e/actions/l2_challenger_test.go
```diff
@@ -165,6 +174,10 @@ interaction:
 		case chal.StatusAsserterTimeout, chal.StatusReadyToProve:
 			txHash = challenger.ActProveFault(t, outputIndex)
 			includeL1Block(t, miner, challenger.address)
+		case chal.StatusProven:
+			// validate l2 out is needed?
```
seolaoh, line 178: I think we need to validate L2 output, and check isValid is true.

e2e/actions/l2_challenger_test.go
```diff
@@ -56,6 +56,15 @@ func TestChallenger(gt *testing.T) {
 		AllowNonFinalized: false,
 	}, miner.EthClient(), challengerRollupClient)
 
+	guardianRPC := e2eutils.NewChallengerL2RPC(proposer.RPCClient())
+	guardianRollupClient := sources.NewRollupClient(guardianRPC)
+	guardian := NewL2Validator(t, log, &ValidatorCfg{
+		OutputOracleAddr:    sd.DeploymentsL1.L2OutputOracleProxy,
+		SecurityCouncilAddr: sd.DeploymentsL1.SecurityCouncilProxy,
+		ValidatorKey:        dp.Secrets.TrustedValidator,
```
seolaoh, line 64: The component `validator` with malicious rpc also has the address of trusted validator, how about change this guardian's `ValidatorKey` to `dp.Secrets.Challenger` as in `e2e/setup.go`?
I think it's weird the malicious validator and guardian have the same `ValidatorKey` even though it's test case.

e2e/system_test.go
```diff
@@ -1769,6 +1772,79 @@ func TestStopStartBatcher(t *testing.T) {
...
+	cfg := DefaultSystemConfig(t)
+
+	sys, err := cfg.Start()
+	require.NoError(t, err, "Error starting up system")
+	defer sys.Close()
+
+	l1Client := sys.Clients["l1"]
+
+	// deposit to ValidatorPool to be a validator / challenger
+	err = cfg.DepositValidatorPool(l1Client, cfg.Secrets.TrustedValidator, big.NewInt(1_000_000_000))
```
Pangssu, line 1790: At this point, the trusted validator has already made a deposit into the ValidatorPool.

Pangssu: in cfg.Start()

e2e/setup.go
```diff
@@ -123,7 +125,7 @@ func DefaultSystemConfig(t *testing.T) SystemConfig {
 		ColosseumSegmentsLengths:  "3,3",
 
 		SecurityCouncilNumConfirmationRequired: 1,
-		SecurityCouncilOwners:                  []common.Address{addresses.Alice, addresses.Bob, addresses.Mallory},
+		SecurityCouncilOwners:                  []common.Address{addresses.Challenger, addresses.Alice, addresses.Bob, addresses.Mallory},
```
Pangssu, line 128: Why is the challenger a member of the security council here?

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -385,6 +378,23 @@ contract Colosseum is Initializable, Semver {
...
 
+    /**
+     * @notice Calls SecurityCouncil to request validation of output to replace with.
+     *
+     * @param _outputIndex Index of the L2 checkpoint output.
+     * @param _outputRoot  The L2 output root to replace the existing one.
+     */
+    function _callSecurityCouncil(uint256 _outputIndex, bytes32 _outputRoot) private {
+        // request outputRoot validation to security council
+        bytes memory CALLDATA = abi.encodeWithSignature("approveChallenge(uint256)", _outputIndex);
```
Pangssu, line 389: Make sure to name the variable camelCase.
like this.
```suggestion
        bytes memory data = abi.encodeWithSignature("approveChallenge(uint256)", _outputIndex);
```

Pangssu (changes requested):

e2e/setup.go
kangsorang: I am simulating a situation where the challenger node and the guardian are activated together. To do this, I added challenger to security council owners.

Pangssu: If the challenger sends challenges in tests, the challenger should not be security council.

0xHansLee (commented):
For default setup for e2e, I suggest to set honest validator instead of malicious one.
Then, I suggest to run malicious one in the test case.

0xHansLee (approved):
lgtm

seolaoh (approved):
LGTM

Pangssu (approved):
lgtm

PR #87, opened 2023-06-27
fix(validator): skip finding fault position in asserter timeout

kangsorang:
# Description
When asserter timeout, skip finding fault position since the same segments have been stored in colosseum.


0xHansLee (approved):
LGTM

chokobole (approved):
LGTM

Pangssu (approved):
lgtm

PR #88, opened 2023-06-27
fix(challenger): fetch output with proof when proving fault

0xHansLee:
Output with proof is required when proving fault.


Pangssu (approved):
lgtm

kangsorang (approved):
LGTM

PR #91, opened 2023-06-29
fix(validator): fix incorrect encoding of L2ToL1MessagePasser balance

Pangssu:
# Description
MerkleProof fails because of the difference between the ABI encoding of uint types and bytes types.


0xHansLee (approved):
LGTM

chokobole (approved):
LGTM

PR #92, opened 2023-06-29
feat(validator): apply gas limit multiplier to guardian transactions

Pangssu:
# Description
guardians send transactions from multiple accounts at the same time, so need more than the estimated gas limit.


0xHansLee (approved):

chokobole (approved):
LGTM

PR #95, opened 2023-06-30
feat(validator): estimate the exec gas limit when sending a confirm tx

Pangssu:


components/validator/guardian.go
```diff
@@ -91,16 +91,60 @@ func (g *Guardian) Stop() error {
...
 func (g *Guardian) ConfirmTransaction(ctx context.Context, transactionId *big.Int) (*types.Transaction, error) {
-	txOpts := utils.NewSimpleTxOpts(ctx, g.cfg.TxManager.From(), g.cfg.TxManager.Signer)
+	cCtx, cCancel := context.WithTimeout(ctx, g.cfg.NetworkTimeout)
+	executionTx, err := g.securityCouncilContract.Transactions(&bind.CallOpts{Context: cCtx}, transactionId)
+	cCancel()
+	if err != nil {
+		return nil, err
+	}
+	if executionTx.Executed {
+		return nil, fmt.Errorf("transaction %d already executed", transactionId)
```
kangsorang, line 111: If the tx was already executed, it would be nice to return err = nil to terminate the loop.
And I think we can conditionally avoid sending a sendTrasaction for a tx that is nil.
```suggestion
		return nil, nil
```

components/validator/guardian.go
```diff
@@ -91,16 +91,60 @@ func (g *Guardian) Stop() error {
...
+		Data:  confirmTx.Data(),
+		Value: confirmTx.Value(),
+	})
+	cCancel()
+	if err != nil {
+		return nil, fmt.Errorf("failed to estimate confirmation tx gas: %w", err)
+	}
+
+	txOpts.Context = ctx
+	txOpts.GasLimit = (confirmationGas * GAS_LIMIT_MULTIPLIER / GAS_LIMIT_DENOMINATOR) + executionGas
```
seolaoh, line 147: We need to use camel case instead of snake case for const variables.

components/validator/guardian.go
```diff
@@ -134,31 +178,35 @@ func (g *Guardian) processOutputValidation(ctx context.Context, event *bindings.
...
 				cCtx, cCancel := context.WithTimeout(ctx, g.cfg.NetworkTimeout)
 				tx, err := g.ConfirmTransaction(cCtx, event.TransactionId)
 				cCancel()
 				if err != nil {
-					g.log.Error("tx call ConfirmTransaction failed", "err", err, "transactionId", event.TransactionId)
+					g.log.Error("failed to create confirm tx", "transactionId", event.TransactionId, "err", err)
 					break Loop
 				}
 
 				g.sendTransaction(tx)
```
seolaoh, line 207: Could you add error handling?

components/validator/guardian.go
```diff
@@ -134,31 +181,39 @@ func (g *Guardian) processOutputValidation(ctx context.Context, event *bindings.
...
 				cCtx, cCancel := context.WithTimeout(ctx, g.cfg.NetworkTimeout)
 				tx, err := g.ConfirmTransaction(cCtx, event.TransactionId)
 				cCancel()
 				if err != nil {
-					g.log.Error("tx call ConfirmTransaction failed", "err", err, "transactionId", event.TransactionId)
+					g.log.Error("failed to create confirm tx", "transactionId", event.TransactionId, "err", err)
 					break Loop
 				}
+				if tx == nil {
+					log.Error("failed to send confirm tx", "transactionId", event.TransactionId, "err", err)
```
seolaoh, line 210:
```suggestion
					g.log.Error("confirm tx is nil", "transactionId", event.TransactionId, "err", err)
```

components/validator/guardian.go
```diff
@@ -91,16 +93,61 @@ func (g *Guardian) Stop() error {
...
+		Data:  confirmTx.Data(),
+		Value: confirmTx.Value(),
+	})
+	cCancel()
+	if err != nil {
+		return nil, fmt.Errorf("failed to estimate confirmation tx gas: %w", err)
+	}
+
+	txOpts.Context = ctx
+	txOpts.GasLimit = (confirmationGas * GasLimitMultiplier / GasLimitDenominator) + executionGas
```
kangsorang, line 150: I think it would be clearer if we declared just one GasLimitMultiplier as a float type.
```suggestion
    const (
        GasLimitMultiplier float32 = 1.5
    )
	txOpts.GasLimit = uint64(float32(confirmationGas) * GasLimitMultiplier) + executionGas
```

seolaoh (approved):
LGTM

PR #99, opened 2023-07-04
perf(validator): remove delay of output handling

seolaoh:
# Description

When challenger handles each output, it should handle it without any delay. Therefore, changed to listen `ticker.C` in post-condition of for loop.
Also, during challenge handling, the same logic is applied.

### DIFF
--- a/components/validator/challenger.go
+++ b/components/validator/challenger.go
@@ -302,14 +302,15 @@ func (c *Challenger) handleOutput(ctx context.Context, outputIndex *big.Int) {
 	ticker := time.NewTicker(time.Minute)
 	defer ticker.Stop()
 
-	for {
-	Loop:
+	for ; ; <-ticker.C {
 		select {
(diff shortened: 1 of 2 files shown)


chokobole:
How about trying this [trick](https://gosamples.dev/range-over-ticker/)? See `Run time.Ticker loop with immediate start!` section

0xHansLee:
> How about trying this [trick](https://gosamples.dev/range-over-ticker/)? See Run time.Ticker loop with immediate start! section

Since we need to handle the case of `ctx.Done()`, I think select case statement is appropriate for our case.

chokobole (approved):
LGTM

0xHansLee (approved):
LGTM

0xHansLee:
Please update the description of this PR. We now use `time.Ticker` again.

Pangssu (approved):
LGTM

PR #104, opened 2023-07-06
feat(contracts): unbond multiple bonds

### LINKED ISSUE #94: Unexpected pending bonds should be released after finalization
### Issue Type

Bug

### Kroma Version

0.2.0

### Current Behaviour?

When an output is submitted late, there are bonds that are pending even though finalization has passed.

### Standalone code to reproduce the issue

If no output is submitted for a certain period of time, a pending bond is created.
Turning trusted validator off for a while is one way to reproduce this issue.

### Expected Behaviour

The pending bonds should be released appropriately.

seolaoh:
# Description

The submission round is split into public round and priority round, and previously at most one bond is unbonded when submitting each output. Therefore, bonds to be unbonded could be pending when output submission have beed delayed. So changed to unbond multiple bonds up to `MAX_UNBOND` in each output submission, and select next priority validator using the last unbonded output.

Fixes #94


kangsorang:
Instead of setting and using the MAX_UNBOND value, it would be nice to attach a counter to the bonds map and use that value.

packages/contracts/contracts/L1/ValidatorPool.sol
```diff
@@ -249,25 +256,40 @@ contract ValidatorPool is ReentrancyGuardUpgradeable, Semver {
...
+        for (; i < MAX_UNBOND; ) {
+            outputIndex = nextUnbondOutputIndex + i;
+            bond = bonds[outputIndex];
+            bondAmount = bond.amount;
 
-        Types.Bond storage bond = bonds[outputIndex];
-        uint128 bondAmount = bond.amount;
-        if (block.timestamp >= bond.expiresAt && bondAmount > 0) {
-            delete bonds[outputIndex];
+            if (block.timestamp >= bond.expiresAt && bondAmount > 0) {
```
kangsorang, line 270: There are no cases where the amount value is 0 on the bond map. 
```suggestion
            if (block.timestamp >= bond.expiresAt) {
```

packages/contracts/contracts/L1/ValidatorPool.sol
```diff
@@ -249,25 +256,40 @@ contract ValidatorPool is ReentrancyGuardUpgradeable, Semver {
...
+
+                unchecked {
+                    ++i;
+                }
+            } else {
+                break;
             }
+        }
+
+        if (i > 0 && output.timestamp > 0) {
```
kangsorang, line 288: Here, i > 0 and output.timestamp > 0 seem like equivalent levels of comparison.
Wouldn't one conditional comparison be enough?

packages/contracts/contracts/L1/ValidatorPool.sol
```diff
@@ -249,25 +256,40 @@ contract ValidatorPool is ReentrancyGuardUpgradeable, Semver {
      * @return Whether the bond has been successfully unbonded.
      */
     function _tryUnbond() private returns (bool) {
-        uint256 outputIndex = nextUnbondOutputIndex;
+        uint256 outputIndex;
+        uint128 bondAmount;
+        Types.Bond storage bond;
+        Types.CheckpointOutput memory output;
+
+        uint256 i = 0;
```
chokobole, line 264: What about renaming this to `unbonded` for readability?

packages/contracts/contracts/L1/ValidatorPool.sol
```diff
@@ -249,25 +256,40 @@ contract ValidatorPool is ReentrancyGuardUpgradeable, Semver {
...
+                unchecked {
+                    ++i;
+                }
+            } else {
+                break;
             }
+        }
+
+        if (i > 0 && output.timestamp > 0) {
+            nextUnbondOutputIndex = nextUnbondOutputIndex + i;
```
chokobole, line 289:
```suggestion
            unchecked {
                        nextUnbondOutputIndex = nextUnbondOutputIndex + i;
            }      
```

packages/contracts/contracts/L1/ValidatorPool.sol
```diff
@@ -249,25 +256,40 @@ contract ValidatorPool is ReentrancyGuardUpgradeable, Semver {
...
     function _tryUnbond() private returns (bool) {
-        uint256 outputIndex = nextUnbondOutputIndex;
+        uint256 outputIndex;
+        uint128 bondAmount;
+        Types.Bond storage bond;
+        Types.CheckpointOutput memory output;
+
+        uint256 i = 0;
+        for (; i < MAX_UNBOND; ) {
+            outputIndex = nextUnbondOutputIndex + i;
```
chokobole, line 266:
```suggestion
            if (outputIndex == 0) {
                outputIndex = nextUnbondOutputIndex;
            }
```

seolaoh:
> Instead of setting and using the MAX_UNBOND value, it would be nice to attach a counter to the bonds map and use that value.

`MAX_UNBOND` is to prevent an unexpected delay of unbonds from causing a single user to use a lot of gas trying to process all of the delayed unbonds. Therefore, a counter of bonds map cannot be used as `MAX_UNBOND`, we need to set it appropriately.

packages/contracts/contracts/L1/ValidatorPool.sol
seolaoh: I think this logic is needed to check bond existence. If the bond of specific `outputIndex` does not exist, `bond.expiresAt` and `bond.amount` is initialized to 0.

packages/contracts/contracts/L1/ValidatorPool.sol
seolaoh: Yes, it's redundant condition. I'll remove it.

chokobole (approved):
LGTM

0xHansLee (approved):
lgtm
Please update the commit msg: `feat(contracts): unbond multiple bonds`

packages/contracts/contracts/L1/ValidatorPool.sol
```diff
@@ -243,31 +250,51 @@ contract ValidatorPool is ReentrancyGuardUpgradeable, Semver {
...
+                }
+            } else {
+                break;
+            }
+        }
 
+        if (unbondedNum > 0) {
             unchecked {
-                ++nextUnbondOutputIndex;
+                nextUnbondOutputIndex += unbondedNum;
```
Pangssu, line 293:
```suggestion
nextUnbondOutputIndex = outputIndex;
```

PR #89, opened 2023-06-27
fix(chain-ops): fix constructor args for validator reward vault

Pangssu:
# Description

This PR includes the fixes below.
- Added a hardhat task to migrate predeployed L2 contracts to a deployment of `hardhat-deploy`.
- Modify `deploy-utils` code to allow L2 contracts to be upgraded.
- Fixed an issue where the constructor argument of the `ValidatorRewardVault` contract was set to an incorrect value when creating an L2 genesis in `chain-ops`.
- Added `--verifier-url` flag to `forge-verify` script.

PR #106, opened 2023-07-06
chore(contracts): remove wrong bytecode diff checking logic

Pangssu:
# Description

Remove the bytecode check logic because the conditional statement checking for contract changes was always comparing the same value, and the hardhat-deploy package already checks for changes.

Related: #89, 369728d

### DIFF
--- a/packages/contracts/src/deploy-utils.ts
+++ b/packages/contracts/src/deploy-utils.ts
@@ -40,26 +40,9 @@ interface DeployOptions {
 export const deploy = async (
   hre: HardhatRuntimeEnvironment,
   name: string,
-  opts?: DeployOptions
+  opts: DeployOptions = {}
 ): Promise<Contract | null> => {
-  if (!opts) {
-    opts = {}
-  }
-
   const { deployer } = await hre.getNamedAccounts()
-  const deployment = await hre.deployments.getOrNull(name)
-
-  if (deployment?.address) {
-    const deployedBytecode = await hre.ethers.provider.getCode(
-      deployment.address
-    )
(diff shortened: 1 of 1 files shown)


0xHansLee (approved):
LGTM

FYI, this is the part that `hardhat-deploy` compares the compiled code and the deployed one as @Pangssu said.
https://github.com/wighawag/hardhat-deploy/blob/1a216e1584fd10262b365fa0469d099cd094ead4/src/helpers.ts#L1013C13-L1013C13

chokobole (approved):
LGTM

kangsorang (approved):
LGTM

packages/contracts/src/deploy-utils.ts
```diff
@@ -97,7 +70,22 @@ export const deploy = async (
...
 
-  if (result.newlyDeployed && typeof opts.postDeployAction === 'function') {
+  if (result.newlyDeployed) {
+    // Always wait for the transaction to be mined, just in case.
+    await hre.ethers.provider.waitForTransaction(result.transactionHash)
+    console.log(`deployed ${name} at ${result.address}`)
+  } else {
+    // If the contract is not newly deployed, do not proceed further.
+    return created
+  }
```
seolaoh, line 80:
```suggestion
  // If the contract is not newly deployed, do not proceed further.
  if (!result.newlyDeployed) {
    return created
  }

  // Always wait for the transaction to be mined, just in case.
  await hre.ethers.provider.waitForTransaction(result.transactionHash)
  console.log(`deployed ${name} at ${result.address}`)
```

seolaoh (approved):
LGTM

PR #114, opened 2023-07-11
feat!(validator): rename output submitter and challenger flags

0xHansLee:
Renamed two validator flags for consistency and readability: 
`VALIDATOR_OUTPUT_SUBMITTER_ENABLED` and `VALIDATOR_CHALLENGER_ENABLED`, both are required.

At first, we named the flags with `*_DISABLED`. If no value is set, it was to turn on the output submitter and challenger by default.

With this renaming, we should provide a sample `.env` file with default value of `true` for both flags for guide of running Kroma node.


chokobole (changes requested):
I think this should be committed with the commit type 'feat' and it should support all old flags for backward compatibility. Additionally, the help message should indicate that they are going to be deprecated.

0xHansLee:
> I think this should be committed with the commit type feat and it should support all old flags for backward compatibility. Additionally, the help message should indicate that they are going to be deprecated.

Right. Thank you for pointing it out. We should support backward compatibility. I'll restore the old flags and mark `deprecated` as you suggested.

0xHansLee:
As we discussed offline, these changes will be included in the next release without backward compatibility.
We should keep in mind that the guides would also be provided accordingly, so I added `!` in the commit.

Pangssu (approved):
LGTM

components/validator/flags/flags.go
```diff
@@ -68,10 +62,11 @@ var (
...
 	}
-	OutputSubmitterDisabledFlag = cli.BoolFlag{
-		Name:   "output-submitter.disabled",
-		Usage:  "Disable l2 output submitter",
-		EnvVar: kservice.PrefixEnvVar(envVarPrefix, "OUTPUT_SUBMITTER_DISABLED"),
+	OutputSubmitterEnabledFlag = cli.BoolFlag{
+		Name:     "output-submitter.enabled",
+		Usage:    "Enable l2 output submitter",
+		EnvVar:   kservice.PrefixEnvVar(envVarPrefix, "OUTPUT_SUBMITTER_ENABLED"),
+		Required: true,
```
seolaoh, line 69: Please move the required flags above `// Optional flags`.

components/validator/flags/flags.go
```diff
@@ -91,10 +86,17 @@ var (
...
+	ChallengerEnabledFlag = cli.BoolFlag{
+		Name:     "challenger.enabled",
+		Usage:    "Enable challenger",
+		EnvVar:   kservice.PrefixEnvVar(envVarPrefix, "CHALLENGER_ENABLED"),
+		Required: true,
+	}
+	ProverGrpcFlag = cli.StringFlag{
+		Name:     "prover-grpc-url",
+		Usage:    "gRPC URL for kroma-prover.",
+		Required: false,
```
seolaoh, line 98: Don't need this line because `Required` default value is `false`.

components/validator/flags/flags.go
```diff
@@ -126,11 +128,11 @@ var requiredFlags = []cli.Flag{
 
 var optionalFlags = []cli.Flag{
 	AllowNonFinalizedFlag,
-	OutputSubmitterDisabledFlag,
+	OutputSubmitterEnabledFlag,
```
seolaoh, line 131: Please move these required flags to `requiredFlags`, and move `ProverGrpcFlag` to here.

components/validator/config.go
```diff
@@ -144,11 +144,11 @@ func NewCLIConfig(ctx *cli.Context) CLIConfig {
 
 		// Optional Flags
 		AllowNonFinalized:            ctx.GlobalBool(flags.AllowNonFinalizedFlag.Name),
-		OutputSubmitterDisabled:      ctx.GlobalBool(flags.OutputSubmitterDisabledFlag.Name),
+		OutputSubmitterEnabled:       ctx.GlobalBool(flags.OutputSubmitterEnabledFlag.Name),
```
seolaoh, line 147: Please move these required flags above `// Optional Flags`, and move `ProverGrpc` here.

components/validator/config.go
```diff
@@ -189,11 +189,11 @@ func NewValidatorConfig(cfg CLIConfig, l log.Logger, m metrics.Metricer) (*Confi
...
 	}
 
-	if cfg.OutputSubmitterDisabled && cfg.ChallengerDisabled {
+	if !cfg.OutputSubmitterEnabled && !cfg.ChallengerEnabled {
 		return nil, errors.New("output submitter and challenger are disabled. either output submitter or challenger must be enabled")
 	}
 
-	if !cfg.ChallengerDisabled && len(cfg.ProverGrpc) == 0 {
+	if cfg.ChallengerEnabled && len(cfg.ProverGrpc) == 0 {
 		return nil, errors.New("ProverGrpc is required but given empty")
```
seolaoh, line 197:
```suggestion
		return nil, errors.New("ProverGrpc is required when challenger enabled, but given empty")
```

e2e/setup.go
```diff
@@ -607,7 +607,8 @@ func (cfg SystemConfig) Start(_opts ...SystemConfigOption) (*System, error) {
 		OutputSubmitterBondAmount:    cfg.OutputSubmitterBondAmount,
 		OutputSubmitterRetryInterval: 50 * time.Millisecond,
 		OutputSubmitterRoundBuffer:   30,
-		ChallengerDisabled:           true,
+		ChallengerEnabled:            false,
+		OutputSubmitterEnabled:       true,
```
seolaoh, line 611: We can remove `ProverGrpc` of asserter.

ops-devnet/docker-compose.yml
```diff
@@ -122,7 +122,7 @@ services:
       VALIDATOR_METRICS_ENABLED: "true"
       VALIDATOR_ALLOW_NON_FINALIZED: "true"
       VALIDATOR_OUTPUT_SUBMITTER_BOND_AMOUNT: 1
-      VALIDATOR_CHALLENGER_DISABLED: "true"
+      VALIDATOR_CHALLENGER_ENABLED: "false"
```
seolaoh, line 125: We need to add `VALIDATOR_OUTPUT_SUBMITTER_ENABLED` to `kroma-validator` also.

ops-devnet/docker-compose.yml
```diff
@@ -151,7 +151,7 @@ services:
       VALIDATOR_COLOSSEUM_ADDRESS: "${COLOSSEUM_ADDRESS}"
       VALIDATOR_VALPOOL_ADDRESS: "${VALPOOL_ADDRESS}"
       VALIDATOR_PROVER_GRPC: "${PROVER_GRPC}"
-      VALIDATOR_OUTPUT_SUBMITTER_DISABLED: "true"
+      VALIDATOR_OUTPUT_SUBMITTER_ENABLED: "false"
```
seolaoh, line 154: We need to add `VALIDATOR_CHALLENGER_ENABLED` to `kroma-challenger` also.

0xHansLee:
@seolaoh Thank you. I applied all your comments. Please review again.

kangsorang (approved):
lgtm

seolaoh (approved):
LGTM

chokobole (approved):
LGTM

Issue #116, opened 2023-07-12
Prevent further progress in the challenge process for finalized output


seolaoh:
### Issue Type

Performance

### Kroma Version

0.2.1

### Current Behaviour?

Currently, a challenger continues challenge process for an output that have already been finalized. Also a guardian processes a validation request of the already finalized output.
For outputs that have already been finalized, this is unnecessary actions and therefore consumes gas unnecessarily.

### Standalone code to reproduce the issue

Run a challenger or guardian node, if an error occurred while progressing challenge process, validation request respectively, it loops forever.

### Expected Behaviour

 Revert challenge process of already finalized output in Colosseum contract, and prevent sending unnecessary transactions in each component (challenger, guardian).

PR #119, opened 2023-07-13
feat: prevent progressing challenge of finalized output

seolaoh:
# Description

We don't need to progress challenge process further when related output has already finalized. 
Therefore, revert challenge process of finalized output in `Colosseum`. Also, prevent `challenger` and `guardian` sending unnecessary transactions which are related to challenge of finalized output.

Related #116


chokobole (approved):
LGTM

components/validator/guardian.go
```diff
@@ -176,12 +190,21 @@ func (g *Guardian) processOutputValidation(ctx context.Context, event *bindings.
 	Loop:
 		select {
 		case <-ticker.C:
-			cCtx, cCancel := context.WithTimeout(ctx, g.cfg.NetworkTimeout)
-			callOpts := utils.NewCallOptsWithSender(cCtx, g.cfg.TxManager.From())
-			isConfirmed, err := g.securityCouncilContract.IsConfirmed(callOpts, event.TransactionId)
-			cCancel()
+			outputIndex := new(big.Int).Sub(event.L2BlockNumber, g.submissionInterval)
```
0xHansLee, line 193: I think that the outputIndex should be obtained via `l2ooContract.getL2OutputIndexAfter()`.

0xHansLee (changes requested):

components/validator/guardian.go
```diff
@@ -34,25 +34,39 @@ type Guardian struct {
 	cancel context.CancelFunc
 	wg     sync.WaitGroup
 
+	submissionInterval *big.Int
+
+	l2ooContract            *bindings.L2OutputOracle
```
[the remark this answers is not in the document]
seolaoh: I didn't add it to `guardian`, because it creates new context with timeout every network request. Otherwise, we can use newly created context with timeout also in `challenger`. Would it be better to use it also in `challenger`?

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -418,6 +423,18 @@ contract Colosseum is Initializable, Semver {
         emit Proven(_outputIndex, _outputRoot);
     }
 
+    /**
+     * @notice Reverts if the output of given index is already finalized.
+     *
+     * @param _outputIndex Index of the L2 checkpoint output.
+     */
+    function _checkOutputNotFinalized(uint256 _outputIndex) private view {
```
kangsorang, line 431: What about changing the function to a modifier?

components/validator/guardian.go
```diff
@@ -176,12 +182,26 @@ func (g *Guardian) processOutputValidation(ctx context.Context, event *bindings.
...
 		select {
 		case <-ticker.C:
-			cCtx, cCancel := context.WithTimeout(ctx, g.cfg.NetworkTimeout)
-			callOpts := utils.NewCallOptsWithSender(cCtx, g.cfg.TxManager.From())
-			isConfirmed, err := g.securityCouncilContract.IsConfirmed(callOpts, event.TransactionId)
-			cCancel()
+			outputIndex, err := g.getL2OutputIndexAfter(ctx, event.L2BlockNumber)
+			if err != nil {
+				g.log.Error("unable to get output index after", "err", err, "l2BlockNumber", event.L2BlockNumber)
+				break Loop
```
kangsorang, line 188: Clear the Loop, please

components/validator/guardian.go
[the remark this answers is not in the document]
0xHansLee: I see. I think it's out of scope of this PR. Let's keep this way for now.
It seems better to add timeout as you said.

0xHansLee (approved):
LGTM

kangsorang (approved):
LGTM

packages/contracts/contracts/L1/Colosseum.sol
seolaoh: I applied it in last force-push, please check it again! (I also changed the order of some functions in `guardian.go`.)

components/validator/challenger.go
```diff
@@ -135,6 +136,7 @@ func (c *Challenger) initSub(ctx context.Context) {
 
 func (c *Challenger) Start(ctx context.Context) error {
 	c.ctx, c.cancel = context.WithCancel(ctx)
+	c.callOpts = utils.NewSimpleCallOpts(ctx)
```
Pangssu, line 139:
```suggestion
	c.callOpts = utils.NewSimpleCallOpts(c.ctx)
```

Pangssu (changes requested):

Pangssu (approved):
LGTM

PR #120, opened 2023-07-17
fix(node): fix the engine sync api getting stuck

Pangssu:
# Description
The issue where subsequent engine sync API functionalities do not work if an error occurs during the execution of the `BlockRefsWithStatus` function has been fixed.

## Issue
If an error occurs during the execution of the `BlockRefsWithStatus` function in the node's state driver (for example, when trying to query a non-existing block), all subsequent functions of the state driver do not work.
As a result, the node cannot proceed with the derivation, and therefore the geth will not be able to keep up with the latest block.

## Changes
Modified the `BlockRefsWithStatus` function in the state driver to always send a signal to the wait channel (`<- wait`) even when an error occurs.


seolaoh (approved):
LGTM

chokobole (approved):
LGTM

PR #121, opened 2023-07-17
fix(validator): prevent validator unnecessary waiting

seolaoh:
# Description

When first turned on the validator node when the chain is already far along, the validator needs to sync all the L2 blocks before it can attempt to submit output. 
Therefore, changed to polling until sync is completed instead of waiting L2 block time for each block.


chokobole (approved):
LGTM

Pangssu (approved):
LGTM

0xHansLee (approved):
LGTM
This commit will have conflicts with current `dev`, so resolve it carefully.

PR #75, opened 2023-06-21
perf(contracts): remove public input hash in final pair

0xHansLee:
Since the public hash is calculated in proving fault,
we don't need to contain public input hash in final pair anymore.
Accordingly, the validation for this has been removed.
Plus, I separated some validation functions depending on their role.

PR #125, opened 2023-07-20
feat!: add ZK proof verifier to colosseum

0xHansLee:
## Description

ZK proof verifier has been added to colosseum. 
Some related changes related to the prover have also been made:
- passing the trace to the prover
- using prover grpc package instead of git submodule

With this changes, a required flag, `L2_ETH_RPC`, is added to validator.


packages/contracts/contracts/L1/ZKVerifier.sol
```diff
@@ -715,14 +1190,21 @@ contract ZKVerifier {
 
     function verify(
         uint256[] calldata proof,
-        uint256[] calldata target_circuit_final_pair
+        uint256[] calldata target_circuit_final_pair,
+        bytes32 publicInputHash
```
0xHansLee, line 1194: @chokobole 
As we discussed in #75, I modified this `verify` function to use `publicInputHash` as an argument.

seolaoh:
Could you rebase onto `dev` and update `.gas-snapshot`?
Also IMO, the name of PR might be better to be `feat!: add ZK proof verifier to colosseum`, and the third commit message to be `feat(contracts): add zk proof verifier to colosseum`. Because the code coverage of this PR is wider than just validator component, and the third commit covers almost contracts related codes.

0xHansLee:
> Could you rebase onto dev and update .gas-snapshot?
Also IMO, the name of PR might be better to be feat!: add ZK proof verifier to colosseum, and the third commit message to be feat(contracts): add zk proof verifier to colosseum. Because the code coverage of this PR is wider than just validator component, and the third commit covers almost contracts related codes.

Rebased onto `dev` and updated .gas-snapshot. And also I changed title of this PR and commit msg as you suggested. Thx!

e2e/setup.go
```diff
@@ -633,6 +632,7 @@ func (cfg SystemConfig) Start(_opts ...SystemConfigOption) (*System, error) {
 	rpcCl := client.NewBaseRPCClient(cl)
 	validatorMaliciousL2RPC := e2eutils.NewMaliciousL2RPC(rpcCl)
 	validatorCfg.RollupClient = sources.NewRollupClient(validatorMaliciousL2RPC)
+	validatorCfg.L2Client = sys.Clients["proposer"]
```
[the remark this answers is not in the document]
0xHansLee: L2 client set in `NewValidatorConfig` do not have `kroma_getBlockTraceByNumberOrHash` method, which is for getting block trace, so I overwrite `sys.Clients["proposer"]` to `L2Client` (and also for L2Client of challengerCfg`).
I think this is not the right way to figure out, but I don't have any clue for this for now.

components/validator/challenge/fetcher.go
```diff
@@ -82,3 +81,51 @@ func Decode(data []byte) []*big.Int {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to json.Marshal %w", err)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, c.address, bytes.NewReader(reqBytes))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create new reqBody for proof: %w", err)
+	}
+
+	cli := http.Client{}
```
kangsorang, line 114: How about handle timeout handling for HTTP requests at the HTTP protocol level?
```suggestion
	cli := http.Client{Timeout: f.timeout}
```

components/validator/challenge/fetcher.go
```diff
@@ -43,16 +49,14 @@ type ProofAndPair struct {
...
-
-	blockNumberHex := fmt.Sprintf("0x%x", blockNumber)
-	f.logger.Info("received block number hex", "hex", blockNumberHex)
+func (f *Fetcher) FetchProofAndPair(ctx context.Context, trace string) (*ProofAndPair, error) {
+	cCtx, cCancel := context.WithTimeout(ctx, f.timeout)
+	defer cCancel()
 
-	response, err := f.Client.Prove(ctx, &pb.ProofRequest{BlockNumberHex: blockNumberHex})
+	// ProofType_AGG(4) is used to generate proof
+	response, err := f.Client.Prove(cCtx, trace, 4)
```
kangsorang, line 57: Are there any use cases planned for the other proofTypes? 
It would be nice to see them specified as separate types.

components/validator/challenger.go
```diff
@@ -692,23 +694,36 @@ func (c *Challenger) ProveFault(ctx context.Context, outputIndex *big.Int, skipS
...
+		return nil, fmt.Errorf("failed to get block trace(block number: %d): %w", targetBlockNumber.Uint64(), err)
 	}
 
-	proof, err := c.PublicInputProof(ctx, blockNumber)
+	traceBz, err := json.Marshal(trace)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshaling trace(block number: %d): %w", targetBlockNumber.Uint64(), err)
+	}
+
+	fetchResult, err := c.cfg.ProofFetcher.FetchProofAndPair(ctx, string(traceBz))
```
kangsorang, line 721: While waiting for the response of the request to the fetcher, it seems that this validator will be unable to respond to other challenges. The situation where this causes a timeout seems to be a problem.

seolaoh:
Could you change [this redundant code block](https://github.com/kroma-network/kroma/pull/125/files#diff-4167deb7abd18d4dcb0e5483ceacf7040ea71e941bb3505bd1870ea125d50de9R427-R441) like below?
```go
	for _, sysNode := range sys.Nodes {
		err = sysNode.Start()
		if err != nil {
			didErrAfterStart = true
			return nil, err
		}
	}
```

Also [this redundant code block](https://github.com/kroma-network/kroma/pull/125/files#diff-4167deb7abd18d4dcb0e5483ceacf7040ea71e941bb3505bd1870ea125d50de9R457-R471) like below?
```go
	for name, node := range sys.Nodes {
		rpcHandler, err := node.RPCHandler()
		if err != nil {
			didErrAfterStart = true
			return nil, err
		}
		sys.Clients[name] = ethclient.NewClient(rpc.DialInProc(rpcHandler))
	}
	l1Client := sys.Clients["l1"]
```

components/validator/challenge/fetcher.go
```diff
@@ -43,16 +49,14 @@ type ProofAndPair struct {
...
 	if err != nil {
-		f.logger.Warn("could not request", "err", err)
+		f.logger.Warn("could not request fault proof", "err", err)
```
seolaoh, line 59:
```suggestion
		f.logger.Error("could not request fault proof", "err", err)
```

components/validator/challenge/fetcher.go
```diff
@@ -82,3 +81,51 @@ func Decode(data []byte) []*big.Int {
 
 	return result
 }
+
+var _ ProverClient = (*JsonRPCProverClient)(nil)
```
seolaoh, line 85: Why do we need this line?

0xHansLee: This is for explicit implementation of interface in go, which means `JsonRPCProverClient` implements `ProverClient` interface.
ref: https://go.dev/doc/effective_go#blank_implements

components/validator/challenge/fetcher.go
0xHansLee: I wanted to cancel the request when parent context is done.
So I used child context with timeout.

components/validator/challenge/fetcher.go
0xHansLee: I heard that there is no plan for using other proof type, so I hardcoded this as 4.
I'll mark `NOTE` with reference.

seolaoh (approved):
LGTM

Pangssu (approved):
LGTM

kangsorang (approved):
LGTM

packages/contracts/contracts/L1/ZKVerifier.sol
```diff
@@ -715,14 +1190,21 @@ contract ZKVerifier {
...
+            (target_circuit_final_pair[0] >> 136) +
+            ((target_circuit_final_pair[1] & 1) << 136);
         instances[2] = target_circuit_final_pair[2] & ((1 << 136) - 1);
-        instances[3] = (target_circuit_final_pair[2] >> 136) + ((target_circuit_final_pair[3] & 1) << 136);
+        instances[3] =
+            (target_circuit_final_pair[2] >> 136) +
+            ((target_circuit_final_pair[3] & 1) << 136);
 
+        instances[4] = uint256(bytes32(bytes16(publicInputHash)) >> (8 * 16));
+        instances[5] = uint256(bytes32(bytes16(publicInputHash << (8 * 16))) >> (8 * 16));
```
chokobole, line 1207: We can save gas with this code below! If i tested at local, it saves like 30 gas.

```suggestion
        instances[4] = uint256(publicInputHash) >> (8 * 16);
        instances[5] = uint256(publicInputHash) & uint256(2**128 - 1);
```

chokobole (approved):
LGTM

PR #127, opened 2023-07-27
fix: fix vulnerabilities reported to ZKTRIE

jyc228:
# Description

This commit resolves two audit issues.
https://github.com/chainlight-io/2023-kroma-network-audit-issues-client/issues/5
https://github.com/chainlight-io/2023-kroma-network-audit-issues-client/issues/6

Issue KROMA-005 is clearly a critical issue on its face.
However, the vulnerability cannot be exploited because of the following piece of code below.

```solidity
if (currentNode.keyPreimage != bytes32(0)) {
    // NOTE(chokobole): The comparison order is important, because in this setting,
    // first condition is mostly evaluted to be true. When we're sure about
    // database preimage, then we need to enable just one of check below!
    require(
        currentNode.keyPreimage == _key || currentNode.keyPreimage == key,
        "ZKMerkleTrie: invalid key preimage"
    );
}
```

Note that we're only checking if currentNode.keyPreimage exists, but @chokobole told me it always exists, so it's not a vulnerability as a result.
However, it's definitely a problem in the source code, and it's likely to be a problem in the future, so we're fixing it.

### DIFF
--- a/packages/contracts/contracts/L1/ZKMerkleTrie.sol
+++ b/packages/contracts/contracts/L1/ZKMerkleTrie.sol
@@ -99,7 +99,10 @@ contract ZKMerkleTrie is IZKMerkleTrie, ZKTrieHasher {
                 );
             } else if (currentNode.nodeType == NodeReader.NodeType.LEAF) {
                 require(!exists && !empty, "ZKMerkleTrie: duplicated terminal node");
-                exists = true;
+                exists = currentNode.nodeKey == key;
+                if (!exists) {
+                    break;
+                }
                 computedKey = _hashFixed3Elems(
                     bytes32(uint256(1)),
                     currentNode.nodeKey,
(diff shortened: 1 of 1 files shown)


packages/contracts/contracts/L1/ZKMerkleTrie.sol
```diff
@@ -99,7 +99,7 @@ contract ZKMerkleTrie is IZKMerkleTrie, ZKTrieHasher {
                 );
             } else if (currentNode.nodeType == NodeReader.NodeType.LEAF) {
                 require(!exists && !empty, "ZKMerkleTrie: duplicated terminal node");
-                exists = true;
+                exists = currentNode.nodeKey == _key;
```
chokobole, line 102: If it doesn't exist, you can break the loop early. Plus, thanks to adding this line, we can safely remove the if condition below `if (currentNode.keyPreimage...`)

jyc228: I will process it as a require to maintain uniformity with other codes.
Shouldn't we think more about removing the verification part with keyPrimage? If that part remains, it can also be verified in the event of a hash collision.

Pangssu (approved):
LGTM

0xHansLee (approved):
LGTM

PR #132, opened 2023-08-01
feat(contracts): add function for forcefully deleting output root

Pangssu:
# Description

If a ZK proof is not possible, the Security Council will force the output root to be deleted by consensus.


packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -882,6 +897,19 @@ contract Colosseum is Initializable, Semver {
...
     }
 
+    /**
+     * @notice Determines whether the given status is ready to be proven.
+     *
+     * @param _status Current status of the challenge.
+     *
+     * @return Whether the given status is ready to be proven.
+     */
+    function _isReadyToProve(ChallengeStatus _status) internal view returns (bool) {
```
seolaoh, line 907: Why not `private pure`?

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -569,6 +569,21 @@ contract Colosseum is Initializable, Semver {
         emit ChallengeDismissed(_outputIndex, _challenger, block.timestamp);
     }
 
+    /**
+     * @notice Deletes the L2 output root forcefully by the Security Council
+     *         when zk-proofing is not possible due to a bug.
```
seolaoh, line 574:
```suggestion
     *         when zk-proving is not possible due to a bug.
```

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -569,6 +569,21 @@ contract Colosseum is Initializable, Semver {
...
+     *
+     * @param _outputIndex Index of the L2 checkpoint output.
+     */
+    function forceDeleteOutput(uint256 _outputIndex)
+        external
+        onlySecurityCouncil
+        outputNotFinalized(_outputIndex)
+    {
+        // Delete output root.
+        L2_ORACLE.replaceL2Output(_outputIndex, DELETED_OUTPUT_ROOT, SECURITY_COUNCIL);
```
seolaoh, line 584: Don't we need the logic to move pending bonds of the challenges related to the `_outputIndex` to bonds and delete the challenges?

packages/contracts/contracts/L1/Colosseum.sol
Pangssu: This function is only used in one place, so it will be removed.

seolaoh (approved):
LGTM, please update the bindings and snapshots at last before merging!

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -417,7 +429,7 @@ contract Colosseum is Initializable, Semver {
         Types.Challenge storage challenge = challenges[_outputIndex][msg.sender];
         ChallengeStatus status = _challengeStatus(challenge);
 
-        if (_cancelIfOutputDeleted(_outputIndex, challenge.challenger, status)) {
+        if (_cancelIfOutputDeleted(_outputIndex, msg.sender, status)) {
```
0xHansLee, line 432: I think this should be reverted even if there are no problems with the code as of now.
With rollback to the previous one, the below `require` statement (L436-L439) can be removed.
```suggestion
        if (_cancelIfOutputDeleted(_outputIndex, challenge.challenger, status)) {
```

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -569,6 +581,21 @@ contract Colosseum is Initializable, Semver {
         emit ChallengeDismissed(_outputIndex, _challenger, block.timestamp);
     }
 
+    /**
+     * @notice Deletes the L2 output root forcefully by the Security Council
+     *         when zk-proving is not possible due to a bug.
```
0xHansLee, line 586:
```suggestion
     *         when zk-proving is not possible due to an undeniable bug.
```

0xHansLee (approved):
LGTM

kangsorang (approved):
LGTM

PR #134, opened 2023-08-03
fix: validate compressedFlags

jyc228:
# Description

Related https://github.com/chainlight-io/2023-kroma-network-audit-issues-client/issues/8

Depending on the number of parameters in _compressedFlags and _values, _valueHash may not hash even once. 
ex
```solidity
bytes32 valueHash = _valueHash(0, "hello world!")
// valueHash does not hashed. it is "hello world!"
```

Initially, we were going to verify that _compressedFlags is non-zero in _valueHash, but since it's a low-level function, it could be used for anything, so we decided to verify it when creating the leaf node.

### DIFF
--- a/packages/contracts/contracts/libraries/NodeReader.sol
+++ b/packages/contracts/contracts/libraries/NodeReader.sol
@@ -173,6 +173,7 @@ library NodeReader {
             // TODO(chokobole): Do the length check as much as possible at once and read the bytes.
             node.nodeKey = readBytes32(item);
             (uint32 compressedFlags, uint256 valuePreimageLen) = readCompressedFlags(item);
+            require((compressedFlags == 1 && valuePreimageLen == 1) || (compressedFlags == 4 && valuePreimageLen == 4), "NodeReader: invalid compressedFlags");
             node.compressedFlags = compressedFlags;
             node.valuePreimage = new bytes32[](valuePreimageLen);
             for (uint256 i = 0; i < valuePreimageLen; ) {


packages/contracts/contracts/libraries/NodeReader.sol
```diff
@@ -173,6 +173,7 @@ library NodeReader {
             // TODO(chokobole): Do the length check as much as possible at once and read the bytes.
```
0xHansLee, line 173: With this fix, can we remove this `TODO`?

0xHansLee (approved):
lgtm

packages/contracts/contracts/libraries/NodeReader.sol
jyc228: If we add logic at the end to verify that the total number of bytes processed is equal to _proof.length, we can delete the Since the read* functions check for length, I think we can delete them. 
@chokobole Can I delete it?

chokobole: Yes! only if you add the logic to verify the length you read is equal to the given length .

chokobole (approved):
LGTM

Pangssu (approved):
LGTM, you can merge after delete `TODO` line.

jyc228:
Added total length check logic.
Additionally, we fixed the existing test data as it was incorrect.

jyc228:
After talking to @chokobole, I decided to erase Todo at another commit.

PR #154, opened 2023-08-11
fix(contracts): set validator reward scalar to system config contract

Pangssu:


0xHansLee (approved):
LGTM.
Let's keep 2000 of `validatorRewardScalar` for sepolia for backward compatibility.

packages/contracts/contracts/L1/SystemConfig.sol
```diff
@@ -5,30 +5,33 @@ import {
...
-     * @custom:value GAS_CONFIG           Represents an update to txn fee config on L2.
-     * @custom:value GAS_LIMIT            Represents an update to gas limit on L2.
-     * @custom:value UNSAFE_BLOCK_SIGNER  Represents an update to the signer key for unsafe
-     *                                    block distrubution.
+     * @custom:value BATCHER                 Represents an update to the batcher hash.
+     * @custom:value GAS_CONFIG              Represents an update to txn fee config on L2.
+     * @custom:value GAS_LIMIT               Represents an update to gas limit on L2.
+     * @custom:value UNSAFE_BLOCK_SIGNER     Represents an update to the signer key for unsafe
+     *                                       block distribution.
+     * @custom:value VALIDATOR_REWARD_SCALAR Represents and update to validator reward scalar.
```
seolaoh, line 27:
```suggestion
     * @custom:value VALIDATOR_REWARD_SCALAR Represents an update to validator reward scalar.
```

utils/chain-ops/genesis/config.go
```diff
@@ -336,10 +341,11 @@ func (d *DeployConfig) RollupConfig(l1StartBlock *types.Block, l2GenesisBlockHas
...
 			SystemConfig: eth.SystemConfig{
-				BatcherAddr: d.BatchSenderAddress,
-				Overhead:    eth.Bytes32(common.BigToHash(new(big.Int).SetUint64(d.GasPriceOracleOverhead))),
-				Scalar:      eth.Bytes32(common.BigToHash(new(big.Int).SetUint64(d.GasPriceOracleScalar))),
-				GasLimit:    uint64(d.L2GenesisBlockGasLimit),
+				BatcherAddr:           d.BatchSenderAddress,
+				Overhead:              eth.Bytes32(common.BigToHash(new(big.Int).SetUint64(d.GasPriceOracleOverhead))),
+				Scalar:                eth.Bytes32(common.BigToHash(new(big.Int).SetUint64(d.GasPriceOracleScalar))),
+				GasLimit:              uint64(d.L2GenesisBlockGasLimit),
+				ValidatorRewardScalar: eth.Bytes32(common.BigToHash(new(big.Int).SetUint64(d.ValidatorRewardScalar))),
```
seolaoh, line 348: Don't we need to add below line to L466 in this file?
```go
		"validatorRewardScalar": config.ValidatorRewardScalar,
```

packages/contracts/contracts/L1/SystemConfig.sol
```diff
@@ -130,7 +143,8 @@ contract SystemConfig is OwnableUpgradeable, Semver {
         bytes32 _batcherHash,
         uint64 _gasLimit,
         address _unsafeBlockSigner,
-        ResourceMetering.ResourceConfig memory _config
+        ResourceMetering.ResourceConfig memory _config,
```
kangsorang, line 146: If we take the validatorRewardScalar value to be uint64, I guess we could use the slot with gasLimit.
Or maybe leave the 32 bytes unused in case we want to add other config values in the future?

components/node/eth/types.go
```diff
@@ -290,5 +290,7 @@ type SystemConfig struct {
 	Scalar Bytes32 `json:"scalar"`
 	// GasLimit identifies the L2 block gas limit
 	GasLimit uint64 `json:"gasLimit"`
+	// ValidatorRewardScalar identifies the validator reward scalar, and is passed through opaquely to kroma-geth.
+	ValidatorRewardScalar Bytes32 `json:"validatorRewardScalar"`
```
kangsorang, line 294: Wouldn't it be more explicit to take the type as int, or is there another reason?

packages/contracts/contracts/L1/SystemConfig.sol
Pangssu: It's the same concept as the `_scalar` above.
And it's already deployed and used in the L1Block contract as a uint256 type, so it can't be changed to a different type for type matching.

components/node/eth/types.go
Pangssu: Same with `Scalar`

kangsorang (approved):
LGTM

seolaoh (approved):
LGTM

PR #155, opened 2023-08-15
feat(validator): inspect outputs which have undeniable bug

0xHansLee:
# Description

When there is an undeniable bug in a checkpoint output, the guardian will find the output, and validate and handle it.
If it is invalid, guardian will request security council to delete it.


0xHansLee:
I will add some tests for this.

components/validator/guardian.go
```diff
@@ -46,59 +66,282 @@ func NewGuardian(cfg Config, l log.Logger) (*Guardian, error) {
...
+			}
+
+			// currentL1 and creationEndedL1 are used for searching events of ReadyToProve in L1 blocks
+			currentL1 := new(big.Int).SetUint64(status.CurrentL1.Number)
+			creationEndedL1 := new(big.Int).Sub(currentL1, new(big.Int).Div(g.creationPeriodSeconds, big.NewInt(12)))
+
+			g.updateInspectionWindow(creationEndedL1, currentL1)
+
+			// finalizedL2 and creationEndedL2 is used to get outputIndex whose creation period is ended but not finalized
+			finalizedL2 := new(big.Int).Sub(currentBlockNumber, new(big.Int).Div(g.finalizationPeriodSeconds, big.NewInt(2)))
```
kangsorang, line 204:
```suggestion
			finalizedL2 := new(big.Int).Sub(currentBlockNumber, new(big.Int).Div(g.finalizationPeriodSeconds, g.l2BlockTime))
```

components/validator/guardian.go
```diff
@@ -46,59 +66,282 @@ func NewGuardian(cfg Config, l log.Logger) (*Guardian, error) {
...
+				}
+
+				tx, err := g.ForceDeleteOutput(g.ctx, outputIndex)
+				if err != nil {
+					g.log.Error("failed to create tx for output deletion", "err", err, "outputIndex", outputIndex)
+					return true
+				}
+
+				if txResponse := g.cfg.TxManager.SendTransaction(g.ctx, tx); txResponse.Err != nil {
+					g.log.Error("failed to send deletion request tx", "err", err, "outputIndex", outputIndex)
```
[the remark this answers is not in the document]
0xHansLee: good point!

Pangssu (approved):

kangsorang (approved):
lgtm

e2e/actions/l2_challenger_test.go
```diff
@@ -456,3 +456,85 @@ interaction2:
 	require.NoError(rt.t, err)
 	require.Equal(rt.t, balance.Int64(), int64(defaultDepositAmount))
 }
+
+func TestChallengeForceDeleteOutputBySecurityCouncil(t *testing.T) {
```
0xHansLee, line 460: Added E2E action test. Please check this test as well. @Pangssu @kangsorang

components/validator/guardian.go
```diff
@@ -46,59 +67,289 @@ func NewGuardian(cfg Config, l log.Logger) (*Guardian, error) {
...
+						g.log.Error("guardian: failed to get output index after", "err", err, "afterL2Block", creationEndedL2.Uint64())
+						return
+					}
+
+					for i := startOutputIndex; i.Cmp(endOutputIndex) != 1; i.Add(i, common.Big1) {
+						g.wg.Add(1)
+						go g.inspectOutput(new(big.Int).Set(i))
+					}
+
+					g.checkpoint = endOutputIndex
```
seolaoh, line 239: Shouldn't we only check the output up to `endOutputIndex - 1` and update checkpoint with `endOutputIndex - 1`? We get `endOutputIndex` using `GetL2OutputIndexAfter`, so it means `endOutputIndex` can be in creation period yet.

components/validator/guardian.go
```diff
@@ -46,59 +67,289 @@ func NewGuardian(cfg Config, l log.Logger) (*Guardian, error) {
...
+				}()
+			} else {
+				func() {
+					cCtx, cCancel := context.WithTimeout(g.ctx, g.cfg.NetworkTimeout)
+					defer cCancel()
+					outputIndex, err := g.l2ooContract.GetL2OutputIndexAfter(utils.NewSimpleCallOpts(cCtx), creationEndedL2)
+					if err != nil {
+						g.log.Error("guardian: failed to get output index after", "err", err, "afterL2Block", finalizedL2.Uint64())
```
seolaoh, line 247:
```suggestion
						g.log.Error("guardian: failed to get output index after", "err", err, "afterL2Block", creationEndedL2.Uint64())
```

components/validator/guardian.go
```diff
@@ -46,59 +67,289 @@ func NewGuardian(cfg Config, l log.Logger) (*Guardian, error) {
...
+						return
+					}
+
+					for i := new(big.Int).Add(g.checkpoint, common.Big1); i.Cmp(outputIndex) != 1; i.Add(i, common.Big1) {
+						g.wg.Add(1)
+						go g.inspectOutput(outputIndex)
+					}
+
+					g.checkpoint = outputIndex
```
seolaoh, line 256: Shouldn't we only check the output up to `outputIndex - 1` and update checkpoint with `outputIndex - 1`? We get `outputIndex` using `GetL2OutputIndexAfter`, so it means `outputIndex` can be in creation period yet.

components/validator/guardian.go
```diff
@@ -46,59 +67,289 @@ func NewGuardian(cfg Config, l log.Logger) (*Guardian, error) {
...
+					defer cCancel()
+					outputIndex, err := g.l2ooContract.GetL2OutputIndexAfter(utils.NewSimpleCallOpts(cCtx), creationEndedL2)
+					if err != nil {
+						g.log.Error("guardian: failed to get output index after", "err", err, "afterL2Block", finalizedL2.Uint64())
+						return
+					}
+
+					for i := new(big.Int).Add(g.checkpoint, common.Big1); i.Cmp(outputIndex) != 1; i.Add(i, common.Big1) {
+						g.wg.Add(1)
+						go g.inspectOutput(outputIndex)
```
seolaoh, line 253:
```suggestion
						go g.inspectOutput(new(big.Int).Set(i))
```

components/validator/guardian.go
```diff
@@ -231,3 +595,15 @@ func (g *Guardian) isTransactionConfirmed(transactionId *big.Int) (bool, error)
 	defer cCancel()
 	return g.securityCouncilContract.IsConfirmed(utils.NewSimpleCallOpts(cCtx), transactionId)
 }
+
+func (g *Guardian) fetchLatestOutputL2Block(ctx context.Context) (*big.Int, error) {
+	cCtx, cCancel := context.WithTimeout(ctx, g.cfg.NetworkTimeout)
+	defer cCancel()
+	latestBlockNumber, err := g.l2ooContract.LatestBlockNumber(utils.NewSimpleCallOpts(cCtx))
+	if err != nil {
+		g.log.Error("validator unable to get latest block number", "err", err)
```
seolaoh, line 604:
```suggestion
		g.log.Error("guardian unable to get latest block number", "err", err)
```

components/validator/guardian.go
```diff
@@ -189,21 +502,72 @@ func (g *Guardian) checkConfirmCondition(transactionId *big.Int, l2BlockNumber *
...
+	return isValid, nil
 }
 
 func (g *Guardian) ConfirmTransaction(ctx context.Context, transactionId *big.Int) (*types.Transaction, error) {
+	g.log.Info("crafting confirm tx", "transactionId", transactionId)
 	txOpts := utils.NewSimpleTxOpts(ctx, g.cfg.TxManager.From(), g.cfg.TxManager.Signer)
 	return g.securityCouncilContract.ConfirmTransaction(txOpts, transactionId)
 }
 
+func (g *Guardian) ForceDeleteOutput(ctx context.Context, outputIndex *big.Int) (*types.Transaction, error) {
```
seolaoh, line 565: It does not delete output forcefully, just request deletion to security council members.
```suggestion
func (g *Guardian) RequestDeletion(ctx context.Context, outputIndex *big.Int) (*types.Transaction, error) {
```

components/validator/guardian.go
```diff
@@ -46,59 +67,289 @@ func NewGuardian(cfg Config, l log.Logger) (*Guardian, error) {
...
+			currentL1 := new(big.Int).SetUint64(status.CurrentL1.Number)
+			creationPeriodL1 := new(big.Int).Div(g.creationPeriodSeconds, big.NewInt(12))
+			creationPeriodL2 := new(big.Int).Div(g.creationPeriodSeconds, g.l2BlockTime)
+			if currentL1.Cmp(creationPeriodL1) != 1 || currentL2.Cmp(creationPeriodL2) != 1 {
+				g.log.Warn("guardian: there is no output when the creation period is over yet", "currentL1", currentL1, "currentL2", currentL2, "creationPeriodL1", creationPeriodL1, "creationPeriodL2", creationPeriodL2)
+				continue
+			}
+
+			creationEndedL1 := new(big.Int).Sub(currentL1, creationPeriodL1)
+			g.updateInspectionWindow(creationEndedL1, currentL1)
```
seolaoh, line 208: I wonder it does work fine.. here we update inspection window using current L1 info, and below in `inspectOutput` goroutine we use the updated inspection window for all outputs in checkpoint range.
1. is it ok to use the same inspection window for all outputs in checkpoint range?
2. After sending the goroutines and loop back here, update the inspection window again, won't this affect the inspection window value already used in the sent goroutines?

components/validator/guardian.go
```diff
@@ -151,12 +430,46 @@ func (g *Guardian) tryConfirmTransaction(event *bindings.SecurityCouncilValidati
...
+		return fmt.Errorf("failed to get output from L2OutputOracle contract(outputIndex: %d): %w", event.OutputIndex.Uint64(), err)
+	}
+
+	isValid, err := g.ValidateL2Output(g.ctx, output.OutputRoot, output.L2BlockNumber.Uint64())
+	if err != nil {
+		return err
+	}
+
+	if isValid {
+		g.log.Info("output deletion is requested, but it's valid")
```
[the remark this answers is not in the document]
0xHansLee: In general, the log level `Warn` is used for the case of potential issues that may lead to errors or unexpected behavior in the future if not addressed. I think this is not the case, it's normal. Instead, I'll add log of transaction ID and output index.

components/validator/guardian.go
0xHansLee: Right, but in `g.inspectOutput`, it checks if the creation period of the output is ended or not. 
If not, it will loop until the creation period end, then validate it.

components/validator/guardian.go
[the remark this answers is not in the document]
seolaoh: I thought that if one of the security council members requested an output deletion but it wasn't a valid request, guardian should be aware of it and check for possible vulnerabilities. As you said, I agree that the log level being a WARN is not suitable, but it would be good to have guardian be aware of it in some other way such as discord notification, ...

components/validator/guardian.go
seolaoh: In `g.inspectOutput`, if the output is still in creation period, the anonymous function returns false and do not retry.

components/validator/guardian.go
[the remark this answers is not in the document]
0xHansLee: Right. We would have another channel for this. 👍

components/validator/guardian.go
0xHansLee: Oops, that was my mistake. If the output is in creation period, we should return true and retry it, right?

seolaoh: Yes! That would be good :)

components/validator/guardian.go
0xHansLee: Thank you for pointing this out!
As you said, different inspection window should be used for each output.
I made a change for this.

seolaoh (approved):
LGTM

PR #157, opened 2023-08-16
feat(contracts): add check if L1 block hash is non zero

seolaoh:
# Description

When creating challenge or submitting L2 output, there is a check if L1 is not reorged. But getting block hash on L1 contract only works for 256 most recent blocks. So changes to check L1 reorg only for 256 recent blocks.


Pangssu (approved):

0xHansLee (approved):

kangsorang (approved):
LGTM

PR #161, opened 2023-08-18
feat(contracts): apply taxation on challenge rewards

Pangssu:
# Description

Related to issue [[KROMA-015] Asserter can always reclaim bond by self-challenging](https://github.com/chainlight-io/2023-kroma-network-audit-issues-client/issues/15) reported by Theori audit.
Asserters and challengers can collude to exploit the challenge system without any loss. To prevent this, It imposes a tax on challenge rewards.


packages/contracts/contracts/L1/ValidatorPool.sol
```diff
@@ -423,8 +446,10 @@ contract ValidatorPool is ReentrancyGuardUpgradeable, Semver {
         uint256 balance = balances[_validator] + _amount;
 
         if (balance >= REQUIRED_BOND_AMOUNT && !isValidator(_validator)) {
-            validatorIndexes[_validator] = validators.length;
-            validators.push(_validator);
+            if (_validator != SECURITY_COUNCIL) {
```
0xHansLee, line 449: Why don't we add this check to `isValidator` function?
It's optional.

0xHansLee (approved):
LGTM

packages/contracts/contracts/L1/ValidatorPool.sol
Pangssu: When deposit, it can only add a validator if the sender is not already a validator, and it use `!isValidator()` to check this. 
If we add a Security council check in `isValidator()`, the function will returns `false`, and it will be added as a validator.

0xHansLee: Ah, you're right.

packages/contracts/contracts/L1/ValidatorPool.sol
```diff
@@ -297,12 +317,15 @@ contract ValidatorPool is ReentrancyGuardUpgradeable, Semver {
             "ValidatorPool: the output is already finalized"
         );
 
-        uint128 increased = pendingBonds[_outputIndex][_challenger];
+        uint128 pendingBond = pendingBonds[_outputIndex][_challenger];
+        uint128 tax = (pendingBond * TAX_NUMERATOR) / TAX_DENOMINATOR;
+        uint128 increased = pendingBond - tax;
         require(increased > 0, "ValidatorPool: the pending bond does not exist");
```
seolaoh, line 323:
```suggestion
        require(pendingBond > 0, "ValidatorPool: the pending bond does not exist");
        uint128 tax = (pendingBond * TAX_NUMERATOR) / TAX_DENOMINATOR;
        uint128 increased = pendingBond - tax;
```

seolaoh (approved):
LGTM

kangsorang (approved):
LGTM

PR #162, opened 2023-08-20
chore(validator): add detailed info of challenge

0xHansLee:
# Description

Added output index and challenger address to challenger logs when crafting colosseum tx.

### DIFF
--- a/components/validator/challenger.go
+++ b/components/validator/challenger.go
@@ -768,7 +768,7 @@ func (c *Challenger) CreateChallenge(ctx context.Context, outputRange *OutputRan
 }
 
 func (c *Challenger) Bisect(ctx context.Context, outputIndex *big.Int, challenger common.Address) (*types.Transaction, error) {
-	c.log.Info("crafting bisect tx")
+	c.log.Info("crafting bisect tx", "outputIndex", outputIndex, "challenger", challenger)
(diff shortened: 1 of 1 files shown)


Pangssu (approved):

kangsorang (approved):

components/validator/challenger.go
```diff
@@ -813,7 +813,7 @@ func (c *Challenger) CancelChallenge(ctx context.Context, outputIndex *big.Int)
 // ProveFault creates proveFault transaction for invalid output root.
 // TODO: ProveFault will take long time, so that we may have to handle it carefully.
 func (c *Challenger) ProveFault(ctx context.Context, outputIndex *big.Int, challenger common.Address, skipSelectFaultPosition bool) (*types.Transaction, error) {
-	c.log.Info("crafting proveFault tx")
+	c.log.Info("crafting proveFault tx", "outputIndex", outputIndex, "challenger", challenger)
```
seolaoh, line 816: `ProveFault` is only called by challenger, do we need to log challenger address here?

0xHansLee: I think It's also informative as well which could be used related to their challenges.

seolaoh (approved):

PR #163, opened 2023-08-21
feat(validator): prevent proving fault when output is already deleted

seolaoh:
# Description

When handling challenge as a challenger, if output is already deleted by another challenger, prevent unnecessary proving fault. This is because unlike bisecting, proving fault is a process that involves expensive computation.


0xHansLee (approved):
lgtm

kangsorang (approved):
LGTM

PR #166, opened 2023-08-22
fix(validator): prevent challenger restart while sync

seolaoh:
# Description

Challenger continuously restarts when kroma node is syncing. To prevent it, changes to loop when error occurred, not restart. And check sync status in validator initially, because currently all the components in validator commonly check it. We do not log sync status error while syncing because node always does not respond while syncing.

Additionally, since validator is made up of multiple components and can be turned on at the same time, adds component names to the logs so that you can distinguish the logs by component. However, some of the logic in the challenger component works even when only the submitter is enabled, so in that case use `challenge` for the name.


components/validator/challenger.go
```diff
@@ -166,38 +166,8 @@ func (c *Challenger) Start(ctx context.Context) error {
...
-	// if challenge mode on, subscribe L2 output submission events
-	if c.cfg.ChallengerEnabled {
-		c.wg.Add(1)
-		go c.subscribeL2OutputSubmitted()
-	}
-
-	// subscribe challenge creation events
 	c.wg.Add(1)
-	go c.subscribeChallengeCreated()
+	go c.loop()
```
kangsorang, line 170: Wouldn't it be better to put initSub after shouldWaitSyncComplete, since subscribe before shouldWaitSyncComplete is done is meaningless?

components/validator/challenger.go
```diff
@@ -219,6 +189,90 @@ func (c *Challenger) Stop() error {
...
+		c.checkpoint = common.Big1
+	} else {
+		// set checkpoint to latestOutputIndex (nextOutputIndex - 1)
+		c.checkpoint = new(big.Int).Sub(nextOutputIndex, common.Big1)
+	}
+	c.metr.RecordChallengeCheckpoint(c.checkpoint)
+	return nil
+}
+
+func (c *Challenger) shouldWaitSyncComplete() (bool, error) {
```
kangsorang, line 259: Wouldn't it be nice to have the outputSubmitter, challenger, and guardian all apply shouldWaitSyncComplete initially. aa Wouldn't it be nice to have shouldWaitSyncComplete initially in common with the NewValidator?

seolaoh: Good idea. I'll check the sync status initially in `validator.go`, and remove this code blocks from each component.

components/validator/challenger.go
seolaoh: If we move the part to check sync status to `validator.go`, this comment will be resolved automatically.

components/validator/validator.go
```diff
@@ -175,3 +189,64 @@ func (v *Validator) Stop() error {
...
+
+	for ; ; <-ticker.C {
+		select {
+		case <-v.ctx.Done():
+			return
+		default:
+			currentBlockNumber, err := v.fetchCurrentBlockNumber()
+			if err != nil {
+				continue
+			}
```
0xHansLee, line 207: Let's log error here as well.

components/validator/l2_output_submitter.go
```diff
@@ -251,25 +243,25 @@ func (l *L2OutputSubmitter) CalculateWaitTime(ctx context.Context, nextBlockNumb
 	return 0
 }
 
-// hasEnoughDeposit checks if validator has enough deposit to bond when trying output submission.
-func (l *L2OutputSubmitter) hasEnoughDeposit(ctx context.Context) (bool, error) {
+// HasEnoughDeposit checks if validator has enough deposit to bond when trying output submission.
+func (l *L2OutputSubmitter) HasEnoughDeposit(ctx context.Context) (bool, error) {
```
0xHansLee, line 247: I just wonder why you make this function public.

components/validator/validator.go
seolaoh: As I mentioned in PR description, I intended not to log sync status error while syncing because node always does not respond while syncing.

components/validator/challenger.go
```diff
@@ -219,6 +189,64 @@ func (c *Challenger) Stop() error {
...
+	defer ticker.Stop()
+
+	for ; ; <-ticker.C {
+		select {
+		case <-c.ctx.Done():
+			return
+		default:
+			if c.cfg.ChallengerEnabled {
+				if err := c.updateCheckpoint(); err != nil {
+					continue
```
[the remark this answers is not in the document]
seolaoh: Already logs error in L236.

components/validator/l2_output_submitter.go
seolaoh: This function is called in `CalculateWaitTime`, and `CalculateWaitTime` is called in test code. If a function using `ctx` is used in test code, we need to receive `ctx` as an argument. I wanted to mark a function using `ctx` and used in test code as a public function.

components/validator/validator.go
0xHansLee: Then, we need to remove Line 210, right?

components/validator/challenger.go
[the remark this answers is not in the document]
0xHansLee: I suggest to move all the log up to this function. It's enough to just return error from the child function.

components/validator/validator.go
seolaoh: No, after pass `fetchCurrentBlockNumber`, `fetchLatestBlockNumber` must succeed. So if an error occurred, we need to log it.

kangsorang: Is the node very slow to respond during sync, but it does respond?

0xHansLee: I may got your point.
`fetchLatestBlockNumber` is from L1, so if an error occurs, we need to log it.
It's optional, but then we can add `log.Debug` when error occurs in `fetchCurrentBlockNumber`.

0xHansLee (approved):
LGTM

kangsorang:
LGTM, meticulous job

kangsorang (approved):

components/validator/validator.go
seolaoh: I added logging it with debug level.

0xHansLee (approved):
lgtm

Pangssu (approved):

PR #167, opened 2023-08-22
feat(contract): add zero timelock for urgent situation

kangsorang:
# Description

Contract upgrades via governor are subject to a 30-day timelock policy by default. 
However, we may need an urgent contract upgrade. 
For this purpose, we've added a simple protocol that allows us to run a proposal with zero timelock.


packages/contracts/contracts/governance/UpgradeGovernor.sol
```diff
@@ -120,6 +120,41 @@ contract UpgradeGovernor is
...
+     */
+    function queue(
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        bytes32 descriptionHash
+    ) public virtual override returns (uint256) {
+        uint256 proposalId = hashProposal(targets, values, calldatas, descriptionHash);
+
+        require(state(proposalId) == ProposalState.Succeeded, "Governor: proposal not successful");
```
Pangssu, line 135:
```suggestion
        require(state(proposalId) == ProposalState.Succeeded, "UpgradeGovernor: proposal not successful");
```

packages/contracts/contracts/governance/UpgradeGovernor.sol
```diff
@@ -120,6 +120,41 @@ contract UpgradeGovernor is
...
+
+        uint256 delay = _timelock.getMinDelay();
+        // Protocol for reflecting urgent decisions on proposals.
+        // A zero delay is applied if the first element of the argument satisfies conditions.
+        if (targets[0] == address(0) && values[0] == 0 && calldatas[0].length == 0) {
```
Pangssu, line 140: You don't need to check for no calldata.

```suggestion
        if (targets[0] == address(0) && values[0] == 0 && calldatas[0].length == 0) {
```

packages/contracts/contracts/governance/UpgradeGovernor.sol
```diff
@@ -120,6 +120,41 @@ contract UpgradeGovernor is
...
+            delay = 0;
```
Pangssu, line 141: Is the delay always 0 in urgent situations?

packages/contracts/contracts/vendor/GovernorTimelockControlUpgradeable.sol
```diff
@@ -0,0 +1,213 @@
...
+ * executing a Denial of Service attack. This risk will be mitigated in a future release.
+ *
+ * _Available since v4.3._
+ */
+abstract contract GovernorTimelockControlUpgradeable is
+    Initializable,
+    IGovernorTimelockUpgradeable,
+    GovernorUpgradeable
+{
+    // change visibility to override queue method
```
Pangssu, line 32: mark your github name

```suggestion
    // NOTE(kangsorang): change visibility to override queue method
```

packages/contracts/contracts/vendor/GovernorTimelockControlUpgradeable.sol
```diff
@@ -0,0 +1,213 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v4.9.0) (governance/extensions/GovernorTimelockControl.sol)
```
seolaoh, line 2: IMO, we should be able to find the original file exactly with this comment.
```suggestion
// OpenZeppelin Contracts (last updated v4.9.0) (governance/extensions/GovernorTimelockControlUpgradeable.sol)
```

packages/contracts/contracts/governance/UpgradeGovernor.sol
```diff
@@ -120,6 +120,41 @@ contract UpgradeGovernor is
...
+        address[] memory targets,
+        uint256[] memory values,
+        bytes[] memory calldatas,
+        bytes32 descriptionHash
+    ) public virtual override returns (uint256) {
+        uint256 proposalId = hashProposal(targets, values, calldatas, descriptionHash);
+
+        require(state(proposalId) == ProposalState.Succeeded, "UpgradeGovernor: proposal not successful");
+
+        uint256 delay = _timelock.getMinDelay();
```
seolaoh, line 137: Unrelated to this file, could you please rename the constructor argument names of the `UpgradeGovernor` and `TimeLock` contracts in deploy config to the following? The current names don't seem to be intuitive.
```
votingDelayBlocks => governorVotingDelay
votingPeriodBlocks => governorVotingPeriod
proposalThresholdTokens => governorProposalThreshold
votesQuorumFractionPercent => governorQuorumFraction
proposalExecutingLatencySeconds => timeLockMinDelay
```

packages/contracts/contracts/vendor/TimelockControllerUpgradeable.sol
```diff
@@ -0,0 +1,476 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v4.9.0) (governance/TimelockController.sol)
```
seolaoh, line 2: IMO, we should be able to find the original file exactly with this comment.
```suggestion
// OpenZeppelin Contracts (last updated v4.9.0) (governance/TimelockControllerUpgradable.sol)
```

packages/contracts/contracts/vendor/GovernorTimelockControlUpgradeable.sol
```diff
@@ -0,0 +1,213 @@
...
+
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts-upgradeable/governance/GovernorUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/governance/extensions/IGovernorTimelockUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/governance/TimelockControllerUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
```
seolaoh, line 9: I think we need to use our contract, isn't it?
```suggestion
import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";

import "./TimelockControllerUpgradeable.sol";
```

packages/contracts/contracts/governance/UpgradeGovernor.sol
```diff
@@ -120,6 +120,41 @@ contract UpgradeGovernor is
         return super.proposalThreshold();
     }
 
+    /**
+     * @notice Function to queue a proposal to the timelock.
+     *         Added protocol for using custom time-lock zero delay for urgent situations.
+     */
+    function queue(
```
seolaoh, line 127: Isn't it better to place this function above the comment `// The following functions are overridden cause required by Solidity.` in L67?

packages/contracts/contracts/governance/UpgradeGovernor.sol
```diff
@@ -120,6 +120,41 @@ contract UpgradeGovernor is
         return super.proposalThreshold();
     }
 
+    /**
+     * @notice Function to queue a proposal to the timelock.
+     *         Added protocol for using custom time-lock zero delay for urgent situations.
```
seolaoh, line 125: Could you add comments about params and return value?

packages/contracts/contracts/governance/UpgradeGovernor.sol
seolaoh: I heard that we made decision to use `values[0]` as the value of `delay` in day unit, isn't it?

seolaoh: I heard that we made decision to use `values[0]` as the value of I heard that we made decision to use values[0] as the value of delay in day unit, isn't it? in day unit, isn't it?

kangsorang: I want to simplify the rule for emergencies to take only 0 time-locks.
This can be done through the use of calldata, but I expect that simplifying the exceptional behavior will make it clearer.

packages/contracts/contracts/governance/UpgradeGovernor.sol
kangsorang: Yes I do remember :) For the renaming, I'll proceed to the next PR

packages/contracts/contracts/vendor/GovernorTimelockControlUpgradeable.sol
kangsorang: If you look at the contracts under contracts-upgradable in openzeppelin, they all specify non-upgradable, not upgradable. It's a bit weird, but I'll fix it accordingly.

packages/contracts/contracts/vendor/GovernorTimelockControlUpgradeable.sol
kangsorang: I took it lightly since it's used for ABI casting purposes, but I think it's correct to specify it exactly as you said. 
Thank you.

packages/contracts/contracts/governance/UpgradeGovernor.sol
seolaoh: Could you check this comment?

Pangssu (approved):

packages/contracts/deploy/015-UpgradeGovernor.ts
```diff
@@ -41,12 +41,16 @@ const deployFn: DeployFunction = async (hre) => {
 
   // Check variable
   assert((await governor.timelock()) === timeLockProxyAddress)
-  assert((await governor.votingDelay()) === hre.deployConfig.votingDelayBlocks)
   assert(
-    (await governor.votingPeriod()) === hre.deployConfig.votingPeriodBlocks
+    (await governor.votingDelay()).toNumber() ===
```
0xHansLee, line 45: We don't need to check `securityCouncilTokenProxyAddress` and `votesQuorumFractionPercent`?

kangsorang: The current contract structure does not provide getters for either element, so we have not made any modifications to the contract for this check, as it would require modifying the contract. 
Even if it is excluded from the validation, the validity of the token address is validated through the constructor logic, and we believe that the votesQuromFractionPercent will be valid along the same lines if the other values are validated successfully. 

Translated with www.DeepL.com/Translator (free version)

seolaoh (approved):
LGTM

packages/contracts/deploy/015-UpgradeGovernor.ts
0xHansLee: I see. thanks!

0xHansLee (approved):
lgtm

PR #168, opened 2023-08-23
feat(contract): change permissions for securityCouncil

kangsorang:
# Description

Permissions to add, remove members, and modify requirements were granted to the governor contract instead of the multisig wallet itself. 
This allows for unified policy enforcement through the governor.


packages/contracts/deploy/012-SecurityCouncil.ts
```diff
@@ -13,8 +13,13 @@ const deployFn: DeployFunction = async (hre) => {
...
   )
 
+  const UpgradeGovernorProxyAddress = await getDeploymentAddress(
+    hre,
+    'UpgradeGovernorProxy'
+  )
+
   await deploy(hre, 'SecurityCouncil', {
-    args: [ColosseumProxyAddress],
+    args: [ColosseumProxyAddress, UpgradeGovernorProxyAddress],
```
seolaoh, line 22: Please add check that `GOVERNOR` is set correctly in `postDeployAction`.

packages/contracts/contracts/universal/MultiSigWallet.sol
```diff
@@ -158,6 +163,14 @@ abstract contract MultiSigWallet is IMultiSigWallet, ReentrancyGuardUpgradeable
         _;
     }
 
+    /**
+     *
```
seolaoh, line 167:
```suggestion
```

seolaoh (approved):
LGTM

Pangssu:
> > Isn't it easier to check if a member owns an SBT than to edit it via governor?
> 
> There seems to be a slight difference between the two. Allowing this behavior based on SBT ownership alone seems risky because it lacks a vote (or approval) from the members.

I approve for now, but I think it could be improved by making the security council with no member edits, and verifying membership with SBT ownership.

packages/contracts/contracts/universal/MultiSigWallet.sol
```diff
@@ -47,13 +52,13 @@ abstract contract MultiSigWallet is IMultiSigWallet, ReentrancyGuardUpgradeable
     uint256 public transactionCount;
 
     /**
-     * @notice Only allow this contract to call the functions.
-     *         This ensures that function is only executed through a multisig-based process.
+     * @notice Only allow specified governor contract to call the functions.
```
0xHansLee, line 55: AFAIK, the `GOVERNOR` in this contract should be the timelock contract of governor, right?
Since the timelock contract will call this contract.

kangsorang: You're right, thanks for pointing that out. 
I think it's correct to name it governor, but the actual injection should be timelock proxy.

0xHansLee: I got your intention. Then, could you add this in the above comment (L20)?
```suggestion
     * @notice Only allow the specified governor contract to call the functions.
```

packages/contracts/contracts/universal/MultiSigWallet.sol
```diff
@@ -16,6 +16,11 @@ import { IMultiSigWallet } from "./IMultiSigWallet.sol";
  *         Allows multiple parties to agree on transactions before execution.
  */
 abstract contract MultiSigWallet is IMultiSigWallet, ReentrancyGuardUpgradeable {
+    /**
+     * @notice The address of the governor contract. Can be updated via upgrade.
```
0xHansLee, line 20: Why don't we add some comments here about governor's timelock contract?
Would it be okay without any further comments?

kangsorang: From a multi-sig wallet's perspective, I think it's right to say that it don't know if Governor uses time-lock or not.

0xHansLee: I see!

0xHansLee (approved):

Pangssu (approved):

PR #169, opened 2023-08-23
fix(validator): skip request deletion when output already been deleted

0xHansLee:
# Description

If an output has already been deleted, there is no need to make a request to delete the output.
Let guardian skip the deleted output.

### DIFF
--- a/components/validator/guardian.go
+++ b/components/validator/guardian.go
@@ -533,6 +533,11 @@ func (g *Guardian) shouldBeDeleted(outputIndex, fromBlock, toBlock *big.Int) (bo
 		return false, fmt.Errorf("failed to get output from L2OutputOracle contract(outputIndex: %d): %w", outputIndex.Uint64(), err)
 	}
 
+	if IsOutputDeleted(output.OutputRoot) {
+		g.log.Info("output has already been deleted", "outputIndex", outputIndex)
+		return false, nil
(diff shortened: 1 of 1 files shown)


seolaoh (approved):
LGTM

components/validator/guardian.go
```diff
@@ -533,6 +533,11 @@ func (g *Guardian) shouldBeDeleted(outputIndex, fromBlock, toBlock *big.Int) (bo
 		return false, fmt.Errorf("failed to get output from L2OutputOracle contract(outputIndex: %d): %w", outputIndex.Uint64(), err)
 	}
 
+	if IsOutputDeleted(output.OutputRoot) {
```
kangsorang, line 536: This seems like a similar property to conditions like isFinalized above the call. Why not put all of this conditional checking logic in the same place at the same level? (guardian.go:280)

Pangssu (approved):

components/validator/guardian.go
0xHansLee: The function `IsOutputDeleted` is in `challenger.go`.
As you mentioned, these kinds of function is spread all over the validator.
I thought that we are gonna make refactoring for future and arrange them at that time.

kangsorang (approved):
LGTM

PR #176, opened 2023-08-24
feat: add genesis block message

Pangssu:


Pangssu:
close this PR because extra data caanot accept long text.

PR #184, opened 2023-09-04
feat(validator): add flag for try output submitting in public round 

kangsorang:
# Description

Added a flag to determine whether output submission behavior in public rounds is enabled or disabled.

I've included a version upgrade of the urfave cli to set the default value of that flag to false.


kangsorang:
I'll do a separate PR on migrating up the Urfave-cli version.

kangsorang:
I'll do a separate PR on migrating up the Urfave-cli version.

PR #195, opened 2023-09-12
docs: apply markdownlint

seolaoh:
# Description

Since we have markdown linting rule (see [here](https://github.com/kroma-network/kroma/blob/dev/specs/meta/linting.md#markdown)), I recommend to edit markdown files in VS Code and use [this extension](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) to apply markdownlint. With it, you can check if a line should be formatted or not, and also set the default formatter to it and auto-fix on save.

### DIFF
--- a/specs/glossary.md
+++ b/specs/glossary.md
@@ -1,3 +1,4 @@
+<!-- markdownlint-disable MD053 -->
 # Glossary
 
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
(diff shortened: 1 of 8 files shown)


0xHansLee (approved):
lgtm

kangsorang (approved):
LGTM

specs/glossary.md
```diff
@@ -1,3 +1,4 @@
+<!-- markdownlint-disable MD053 -->
```
Pangssu, line 1: Why is MD053 disabled in this file?

seolaoh: MD053 is a rule that `Link and image reference definitions should be needed`.
Without it, we need to remove all unused definitions such as `[storage-root]: glossary.md#storage-root` in this glossary file. But I think it is kind of convention to add the definition of all glossaries to use in future. So I suppressed this rule only for this glossary file.

Pangssu (approved):

PR #198, opened 2023-09-13
feat(service): sending tx with access list by default

Pangssu:
# Description

When sending a transaction using txmgr to save transaction fees, add the access list to the transaction body.


Pangssu:
We decided not to add an access list by default using `createAccessList` api because there are cases where it can be detrimental.

PR #200, opened 2023-09-18
docs: add security audit by ChainLight

0xHansLee:


seolaoh:
How about add a line in `README.md` like below?
`See [here](./audits) for security audit reports.`

0xHansLee:
> How about add a line in README.md like below?
See [here](./audits) for security audit reports.

I also thought about it, but I didn't. I thought we don't need to state here about audits because it is `About Kroma`. 
It'd better to make a separate section for security audits. wdyt?

seolaoh:
> > How about add a line in README.md like below?
> > See [here](./audits) for security audit reports.
> 
> I also thought about it, but I didn't. I thought we don't need to state here about audits because it is `About Kroma`. It'd better to make a separate section for security audits. wdyt?

Yes, I suggested it to add the instruction about audit report somewhere in this repository. It does not need to be in `README.md`.

Pangssu (approved):

kangsorang (approved):
LGTM

seolaoh (approved):
LGTM

PR #223, opened 2023-10-26
feat(validator): add gas limit margin when submitting output

seolaoh:
# Description

Since we have dynamic gas usage in unbond and priority validator selection logic, it is hard to estimate exact gas usage. This leads to frequent transaction failures, so to avoid this, set the output submission gas limit `1.5 * estimated gas`.

### DIFF
--- a/components/validator/l2_output_submitter.go
+++ b/components/validator/l2_output_submitter.go
@@ -10,6 +10,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/ethereum/go-ethereum"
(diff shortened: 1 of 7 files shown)


0xHansLee (approved):
LGTM
But this PR may lead to conflicts with OP upstream PR which is in progress by @kangsorang.
Please handling this appropriately when applying upstream.

Pangssu (approved):

kangsorang (approved):
LGTM

PR #225, opened 2023-11-07
feat: withdraw validator rewards directly to l2

Pangssu:


packages/contracts/contracts/L2/ValidatorRewardVault.sol
```diff
@@ -122,6 +123,24 @@ contract ValidatorRewardVault is FeeVault, Semver {
...
+    function withdrawToL2() external {
+        uint256 balance = rewards[msg.sender];
+        rewards[msg.sender] = 0;
+
+        unchecked {
+            totalReserved -= balance;
+            totalProcessed += balance;
+        }
+
+        emit Withdrawal(balance, msg.sender, msg.sender);
```
seolaoh, line 138: Since this event is from `FeeVault` and it originally means the withdrawal event to L1, how about add new event in this contract that means withdrawal event to L2?

packages/contracts/contracts/L2/ValidatorRewardVault.sol
```diff
@@ -122,6 +123,24 @@ contract ValidatorRewardVault is FeeVault, Semver {
         );
     }
 
+    /**
+     * @notice Withdraws all of the sender's balance to L2.
+     */
+    function withdrawToL2() external {
+        uint256 balance = rewards[msg.sender];
```
seolaoh, line 130: I wonder why not checking if the balance is greater than `MIN_WITHDRAWAL_AMOUNT`?

packages/contracts/contracts/L2/ValidatorRewardVault.sol
Pangssu: The `Withdrawal` event means that a withdrawal request has been made, rather than a withdrawal to L1. 
The event of withdrawing to L1 can be checked as an event emitted by L2 Bridge contract.

packages/contracts/contracts/L2/ValidatorRewardVault.sol
0xHansLee: IMO, the reason why we don't need this check is because the `MIN_WITHDRAWAL_AMOUNT` is 0, not transaction fee. If I'm right, I also think that it's better to check the withdrawal amount for consistency.

0xHansLee (approved):

packages/contracts/contracts/L2/ValidatorRewardVault.sol
seolaoh: `MIN_WITHDRAWAL_AMOUNT` could be a value greater than 0 in the future, so if we decide to check this condition to restrict all withdrawals less than a threshold, it should be added. Otherwise we decide to check this condition because withdrawal fee to L1 is expensive, it could not be added. IMO, the previous semantic seems to be more proper, wdyt?

seolaoh: Oh I see the changes after this comment. I'll check it!

seolaoh (approved):
LGTM

packages/contracts/contracts/L2/ValidatorRewardVault.sol
```diff
@@ -95,33 +96,51 @@ contract ValidatorRewardVault is FeeVault, Semver {
...
     }
 
+    /**
+     * @notice Withdraws all of the sender's balance to L2.
+     *         Reverts if the balance is less than the minimum withdrawal amount.
+     */
+    function withdrawToL2() external {
+        uint256 amount = processWithdrawal();
+
+        bool success = SafeCall.call(msg.sender, gasleft(), amount, bytes(""));
```
kangsorang, line 140: I think it would be more explicit to use the send function in the Safecall library. Is there any other purpose for using call?
```suggestion
        bool success = SafeCall.send(msg.sender, gasleft(), amount);
```

Pangssu: There is currently no `send` function.

kangsorang: Oh, right... I was mistaken :)

kangsorang (approved):
LGTM

PR #235, opened 2023-12-12
fix(contracts): change so that only recipient can withdraw from vault

0xHansLee:
In the previous code, anyone can call `withdraw` or `withdrawToL2` and the reward is sent to the caller. 
This need to be fixed so that the reward is sent to the recipient.

And I added a modifier that only the recipient can call. 
In case of `withdraw`, it requires additional action on L1 (prove and finalize), so this should be controlled only by the recipient.


Pangssu (approved):

kangsorang (approved):
LGTM

seolaoh (approved):
LGTM

PR #247, opened 2024-01-31
fix: add blob related things to rpc header

0xHansLee:
Merge into `release/v1.1.1-base` branch, which is a base branch for releasing `v1.1.1`. 
With dencun upgrade, blob related fields are added to block header. 
Accordingly, those should be added. Without this, the validation of block header when starting node is failed.

```
msg="failed to fetch L1 head for runtime config initialization" err=failed to fetch head header:  failed to verify block hash: computed 0xaaa but RPC said 0xbbb"
```

After merging this PR, `v1.1.1` will be released. Also, this should be applied to `v1.2.0-rc.*`, where the upstreamed version is v1.3.0, since this is added in Optimism `v1.4.0`.

## Summary by CodeRabbit

- **Refactor**
	- Updated the data structure to include new fields for enhanced transaction processing and tracking.

### DIFF
--- a/op-service/sources/types.go
+++ b/op-service/sources/types.go
@@ -136,7 +136,16 @@ type rpcHeader struct {
 	BaseFee *hexutil.Big `json:"baseFeePerGas"`
 
 	// WithdrawalsRoot was added by EIP-4895 and is ignored in legacy headers.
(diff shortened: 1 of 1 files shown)


Pangssu (approved):

Pangssu:
@coderabbitai review

PR #248, opened 2024-01-31
ci: run test ci for every PR

0xHansLee:
Let's run test ci for every PR

## Summary by CodeRabbit

- **Chores**
	- Updated the workflow to trigger for pull requests targeting any branch, enhancing testing coverage.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,10 +1,6 @@
 name: Tests
 
-on:
-  pull_request:
-    branches:
-      - dev
-      - main
+on: pull_request
 
 jobs:
   test:


Pangssu (approved):

.github/workflows/test.yml
```diff
@@ -3,8 +3,7 @@ name: Tests
 on:
   pull_request:
     branches:
-      - dev
-      - main
+      - *
```
seolaoh, line 6: The characters *, [, and ! are special characters in YAML. If you start a pattern with *, [, or !, you must enclose the pattern in quotes. ([link](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet))
How about just changing like below?
```
on: pull_request
```

0xHansLee: Good. I changed as you suggested

seolaoh (approved):

PR #271, opened 2024-03-11
feat(contracts): implement AssetManager

sm-stack:
Implement AssetManager that handles (un)delegation / reward distribution / slashing in governance token.

### DIFF
--- a/packages/contracts/contracts/L1/AssetManager.sol
+++ b/packages/contracts/contracts/L1/AssetManager.sol
@@ -1,0 +1,1627 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.15;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { IERC721 } from "@openzeppelin/contracts/token/ERC721/IERC721.sol";
+import { IERC721Receiver } from "@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol";
+import { Math } from "@openzeppelin/contracts/utils/math/Math.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+import { Atan2 } from "../libraries/Atan2.sol";
+import { BalancedWeightTree } from "../libraries/BalancedWeightTree.sol";
+import { Types } from "../libraries/Types.sol";
+import { Uint128Math } from "../libraries/Uint128Math.sol";
+import { IKGHManager } from "../universal/IKGHManager.sol";
+import { Colosseum } from "./Colosseum.sol";
+import { L2OutputOracle } from "./L2OutputOracle.sol";
+
+/**
+ * @title AssetManager
+ * @notice AssetManager is an abstract contract that handles (un)delegations of KRO and KGH, and
+ *         the distribution of rewards to the delegators and the validator.
+ */
+abstract contract AssetManager is IERC721Receiver {
+    using Math for uint256;
+    using Uint128Math for uint128;
+    using BalancedWeightTree for BalancedWeightTree.Tree;
+
+    /**
+     * @notice Represents the asset information of the vault of a validator.
+     *
+     * @custom:field totalKro                   Total amount of KRO in the vault.
+     * @custom:field totalKroShares             Total shares for KRO delegation in the vault.
+     * @custom:field totalKgh                   Total number of KGH in the vault.
+     * @custom:field totalKroInKgh              Total amount of KRO which KGHs in the vault have.
+     * @custom:field totalKghShares             Total shares for KGH delegation in the vault.
+     * @custom:field validatorKro               Amount of KRO that the validator self-delegated.
+     * @custom:field totalPendingAssets         Total pending KRO for undelegation.
+     * @custom:field totalPendingBoostedRewards Total pending boosted rewards in KRO for undelegation.
+     * @custom:field totalPendingKroShares      Total pending KRO shares for undelegation.
+     * @custom:field totalPendingKghShares      Total pending KGH shares for undelegation.
+     */
+    struct Asset {
+        uint128 totalKro;
+        uint128 totalKroShares;
+        uint128 totalKgh;
+        uint128 totalKroInKgh;
+        uint128 totalKghShares;
+        uint128 validatorKro;
+        uint128 totalPendingAssets;
+        uint128 totalPendingBoostedRewards;
+        uint128 totalPendingKroShares;
+        uint128 totalPendingKghShares;
+    }
+
+    /**
+     * @notice Constructs the reward information of the vault of a validator.
+     *
+     * @custom:field boostedReward                Cumulated boosted reward for KGH delegators in the vault.
+     * @custom:field validatorRewardKro           Cumulated reward for the validator.
+     * @custom:field commissionRate               Commission rate of validator.
+     * @custom:field commissionMaxChangeRate      Maximum changeable commission rate at once.
+     * @custom:field commissionRateChangedAt      Last timestamp when the commission rate was changed.
+     * @custom:field totalPendingValidatorRewards Total pending validator rewards.
+     * @custom:field claimRequestTimes            Timestamps of validator reward claim requests.
+     * @custom:field pendingValidatorRewards      A mapping of timestamp to pending validator rewards.
+     */
+    struct Reward {
+        uint128 boostedReward;
+        uint128 validatorRewardKro;
+        uint8 commissionRate;
+        uint8 commissionMaxChangeRate;
+        uint128 commissionRateChangedAt;
+        uint128 totalPendingValidatorRewards;
+        uint256[] claimRequestTimes;
+        mapping(uint256 => uint128) pendingValidatorRewards;
+    }
+
+    /**
+     * @notice Constructs the delegator of KRO in the vault of a validator.
+     *
+     * @custom:field shares                 Amount of shares for KRO delegation.
+     * @custom:field undelegateRequestTimes Timestamps of undelegation requests.
+     * @custom:field pendingKroShares       A mapping of timestamp undelegations are requested to
+     *                                      pending KRO shares for undelegation.
+     */
+    struct KroDelegator {
+        uint128 shares;
+        uint256[] undelegateRequestTimes;
+        mapping(uint256 => uint128) pendingKroShares;
+    }
+
+    /**
+     * @notice Represents the amounts of KRO and KGH shares that requested for undelegation when undelegating KGH.
+     *
+     * @custom:field kroShares Amount of KRO shares to undelegate.
+     * @custom:field kghShares Amount of KGH shares to undelegate.
+     */
+    struct UndelegateShares {
+        uint128 kroShares;
+        uint128 kghShares;
+    }
+
+    /**
+     * @notice Constructs the delegator of KGH in the vault of a validator.
+     *
+     * @custom:field kroShares              Amount of shares for KRO in KGH with a corresponding tokenId.
+     * @custom:field kghShares              Amount of shares for delegation of KGH with a corresponding tokenId.
+     * @custom:field undelegateRequestTimes Timestamps of undelegation requests.
+     * @custom:field pendingKghIds          A mapping of timestamp undelegations are requested to
+     *                                      token ids of KGHs for undelegation.
+     * @custom:field pendingShares          A mapping of timestamp undelegations are requested to
+     *                                      pending KRO and KGH shares for undelegation.
+     */
+    struct KghDelegator {
+        mapping(uint256 => uint128) kroShares;
+        mapping(uint256 => uint128) kghShares;
+        uint256[] undelegateRequestTimes;
+        mapping(uint256 => uint256[]) pendingKghIds;
+        mapping(uint256 => UndelegateShares) pendingShares;
+    }
+
+    /**
+     * @notice Constructs the vault of a validator.
+     *
+     * @custom:field asset             Asset information of the vault.
+     * @custom:field reward            Reward information of the vault.
+     * @custom:field isInitiated       Whether the vault is initiated.
+     * @custom:field noSubmissionCount Number of counts that the validator did not submit the output in priority round.
+     * @custom:field kroDelegators     A mapping of validator address to KRO delegator struct.
+     * @custom:field kghDelegators     A mapping of validator address to KGH delegator struct.
+     */
+    struct Vault {
+        Asset asset;
+        Reward reward;
+        bool isInitiated;
+        uint8 noSubmissionCount;
+        mapping(address => KroDelegator) kroDelegators;
+        mapping(address => KghDelegator) kghDelegators;
+    }
+
+    /**
+     * @notice The numerator for the boosted reward.
+     */
+    uint128 public constant BOOSTED_REWARD_NUMERATOR = 40;
+
+    /**
+     * @notice The denominator for the boosted reward.
+     */
+    uint128 public constant BOOSTED_REWARD_DENOM = 100;
+
+    /**
+     * @notice The denominator for per-mille used at slashing and initialization of undelegation.
+     */
+    uint128 public constant PER_MILLE_DENOM = 1000;
+
+    /**
+     * @notice Virtual KRO amount per KGH.
+     */
+    uint128 public constant VKRO_PER_KGH = 100e18;
+
+    /**
+     * @notice The numerator of the tax.
+     */
+    uint128 public constant TAX_NUMERATOR = 20;
+
+    /**
+     * @notice The denominator of the tax.
+     */
+    uint128 public constant TAX_DENOMINATOR = 100;
+
+    /**
+     * @notice Decimals offset for the KRO and KGH shares.
+     */
+    uint128 public constant DECIMAL_OFFSET = 10 ** 6;
+
+    /**
+     * @notice Address of the KRO token contract.
+     */
+    IERC20 public immutable ASSET_TOKEN;
+
+    /**
+     * @notice Address of the L2OutputOracle contract. Can be updated via upgrade.
+     */
+    L2OutputOracle public immutable L2_ORACLE;
+
+    /**
+     * @notice Address of the KGH token contract.
+     */
+    IERC721 public immutable KGH;
+
+    /**
+     * @notice Address of the KGHManager contract.
+     */
+    IKGHManager public immutable KGH_MANAGER;
+
+    /**
+     * @notice The address of the SecurityCouncil contract. Can be updated via upgrade.
+     */
+    address public immutable SECURITY_COUNCIL;
+
+    /**
+     * @notice The max number of outputs to be finalized at once.
+     */
+    uint128 public immutable MAX_OUTPUT_FINALIZATIONS;
+
+    /**
+     * @notice Amount of base reward for the validator.
+     */
+    uint128 public immutable BASE_REWARD;
+
+    /**
+     * @notice The numerator of the slashing rate. It will be divided by PER_MILLE_DENOM.
+     */
+    uint128 public immutable SLASHING_RATE_NUMERATOR;
+
+    /**
+     * @notice Minimum amount to slash.
+     */
+    uint128 public immutable MIN_SLASHING_AMOUNT;
+
+    /**
+     * @notice Minimum amount to register as a validator.
+     */
+    uint128 public immutable MIN_REGISTER_AMOUNT;
+
+    /**
+     * @notice Minimum amount to start submitting outputs.
+     */
+    uint128 public immutable MIN_START_AMOUNT;
+
+    /**
+     * @notice Delay for the finalization of undelegation.
+     */
+    uint256 public immutable UNDELEGATION_PERIOD;
+
+    /**
+     * @notice Weighted tree to store and calculate the probability to be selected as an output submitter.
+     */
+    BalancedWeightTree.Tree internal _validatorTree;
+
+    /**
+     * @notice The output index to distribute reward next.
+     */
+    uint256 internal _nextRewardOutputIndex;
+
+    /**
+     * @notice A mapping of validator address to the vault.
+     */
+    mapping(address => Vault) internal _vaults;
+
+    /**
+     * @notice A mapping of the submitted output index to finalization timestamp.
+     */
+    mapping(uint256 => uint256) internal _submittedOutputs;
+
+    /**
+     * @notice A mapping of output index challenged successfully to pending challenge rewards.
+     */
+    mapping(uint256 => uint128) internal _pendingChallengeReward;
+
+    /**
+     * @notice Emitted when KROs are delegated.
+     *
+     * @param validator Address of the validator.
+     * @param delegator Address of the delegator.
+     * @param amount    The amount of KRO delegated.
+     * @param shares    The amount of shares received.
+     */
+    event KroDelegated(
+        address indexed validator,
+        address indexed delegator,
+        uint128 amount,
+        uint128 shares
+    );
+
+    /**
+     * @notice Emitted when a KGH is delegated.
+     *
+     * @param validator Address of the validator.
+     * @param delegator Address of the delegator.
+     * @param tokenId   Token Id of the KGH.
+     * @param kroInKgh  The amount of KRO in the KGH.
+     * @param kroShares The amount of KRO shares received.
+     * @param kghShares The amount of KGH shares received.
+     */
+    event KghDelegated(
+        address indexed validator,
+        address indexed delegator,
+        uint256 tokenId,
+        uint128 kroInKgh,
+        uint128 kroShares,
+        uint128 kghShares
+    );
+
+    /**
+     * @notice Emitted when KGHs are delegated in batch.
+     *
+     * @param validator Address of the validator.
+     * @param delegator Address of the delegator.
+     * @param tokenIds  Array of token ids of the KGHs.
+     * @param kroShares The amount of KRO shares received.
+     * @param kghShares The amount of KGH shares received.
+     */
+    event KghBatchDelegated(
+        address indexed validator,
+        address indexed delegator,
+        uint256[] tokenIds,
+        uint128 kroShares,
+        uint128 kghShares
+    );
+
+    /**
+     * @notice Emitted when KRO undelegation is initiated.
+     *
+     * @param validator Address of the validator.
+     * @param delegator Address of the delegator.
+     * @param amount    The amount of KRO to undelegate.
+     * @param shares    The amount of shares to undelegate.
+     */
+    event KroUndelegationInitiated(
+        address indexed validator,
+        address indexed delegator,
+        uint128 amount,
+        uint128 shares
+    );
+
+    /**
+     * @notice Emitted when KGH undelegation is initiated.
+     *
+     * @param validator Address of the validator.
+     * @param delegator Address of the delegator.
+     * @param tokenId   Token Id of the KGH.
+     * @param kroShares The amount of KRO shares to undelegate.
+     * @param kghShares The amount of KGH shares to undelegate.
+     */
+    event KghUndelegationInitiated(
+        address indexed validator,
+        address indexed delegator,
+        uint256 tokenId,
+        uint128 kroShares,
+        uint128 kghShares
+    );
+
+    /**
+     * @notice Emitted when KGHs undelegation is initiated in batch.
+     *
+     * @param validator Address of the validator.
+     * @param delegator Address of the delegator.
+     * @param tokenIds  Array of token ids of the KGHs.
+     * @param kroShares The amount of KRO shares to undelegate.
+     * @param kghShares The amount of KGH shares to undelegate.
+     */
+    event KghBatchUndelegationInitiated(
+        address indexed validator,
+        address indexed delegator,
+        uint256[] tokenIds,
+        uint128 kroShares,
+        uint128 kghShares
+    );
+
+    /**
+     * @notice Emitted when KRO undelegation is finalized.
+     *
+     * @param validator Address of the validator.
+     * @param delegator Address of the delegator.
+     * @param amount    The amount of KRO undelegated.
+     */
+    event KroUndelegationFinalized(
+        address indexed validator,
+        address indexed delegator,
+        uint128 amount
+    );
+
+    /**
+     * @notice Emitted when KGH undelegation is finalized.
+     *
+     * @param validator Address of the validator.
+     * @param delegator Address of the delegator.
+     * @param amount    The amount of KRO undelegated during the KGH undelegation.
+     */
+    event KghUndelegationFinalized(
+        address indexed validator,
+        address indexed delegator,
+        uint128 amount
+    );
+
+    /**
+     * @notice Emitted when pending challenge reward for challenge winner is distributed.
+     *
+     * @param recipient Address of the reward recipient.
+     * @param amount    The amount of challenge reward.
+     */
+    event ChallengeRewardDistributed(address indexed recipient, uint128 amount);
+
+    /**
+     * @notice Emitted when the validator is slashed.
+     *
+     * @param loser  Address of the loser at the challenge.
+     * @param winner Address of the winner at the challenge.
+     * @param amount The amount of KRO slashed.
+     */
+    event Slashed(address indexed loser, address indexed winner, uint128 amount);
+
+    /**
+     * @notice Emitted when validator reward claim is initiated.
+     *
+     * @param validator Address of the validator.
+     * @param amount    The amount of validator reward requested for claim.
+     */
+    event RewardClaimInitiated(address indexed validator, uint128 amount);
+
+    /**
+     * @notice Emitted when validator reward claim is finalized.
+     *
+     * @param validator Address of the validator.
+     * @param amount    The amount of validator reward claimed.
+     */
+    event RewardClaimFinalized(address indexed validator, uint128 amount);
+
+    /**
+     * @notice Emitted when the output reward is distributed.
+     *
+     * @param recipient Address of the reward recipient.
+     * @param amount    The amount of reward.
+     */
+    event RewardDistributed(address indexed recipient, uint128 amount);
+
+    /**
+     * @notice Modifier to check if the caller is the Colosseum contract.
+     */
+    modifier onlyColosseum() {
+        require(
+            msg.sender == L2_ORACLE.COLOSSEUM(),
+            "AssetManager: Only Colosseum can call this function"
+        );
+        _;
+    }
+
+    /**
+     * @notice Modifier to check if the vault is active.
+     */
+    modifier checkIsActive(address validator) {
+        require(
+            msg.sender == validator || _vaults[validator].asset.validatorKro >= MIN_REGISTER_AMOUNT,
+            "AssetManager: Vault is inactive"
+        );
+        _;
+    }
+
+    /**
+     * @notice Constructs the AssetManager contract.
+     *
+     * @param _l2OutputOracle         Address of the L2OutputOracle contract.
+     * @param _assetToken             Address of the KRO token.
+     * @param _kgh                    Address of the KGH token.
+     * @param _kghManager             Address of the KGHManager contract.
+     * @param _securityCouncil        Address of the SecurityCouncil contract.
+     * @param _maxOutputFinalizations Max number of finalized outputs.
+     * @param _baseReward             Base reward for the validator.
+     * @param _slashingRateNumerator  Numerator of the slashing rate.
+     * @param _minSlashingAmount      Minimum amount to slash.
+     * @param _minRegisterAmount      Minimum amount to register as a validator.
+     * @param _minStartAmount         Minimum amount to start submitting outputs.
+     * @param _undelegationPeriod     Period that should wait to finalize the undelegation.
+     */
+    constructor(
+        L2OutputOracle _l2OutputOracle,
+        IERC20 _assetToken,
+        IERC721 _kgh,
+        IKGHManager _kghManager,
+        address _securityCouncil,
+        uint128 _maxOutputFinalizations,
+        uint128 _baseReward,
+        uint128 _slashingRateNumerator,
+        uint128 _minSlashingAmount,
+        uint128 _minRegisterAmount,
+        uint128 _minStartAmount,
+        uint256 _undelegationPeriod
+    ) {
+        L2_ORACLE = _l2OutputOracle;
+        ASSET_TOKEN = _assetToken;
+        KGH = _kgh;
+        KGH_MANAGER = _kghManager;
+        SECURITY_COUNCIL = _securityCouncil;
+        MAX_OUTPUT_FINALIZATIONS = _maxOutputFinalizations;
+        BASE_REWARD = _baseReward;
+        SLASHING_RATE_NUMERATOR = _slashingRateNumerator;
+        MIN_SLASHING_AMOUNT = _minSlashingAmount;
+        MIN_REGISTER_AMOUNT = _minRegisterAmount;
+        MIN_START_AMOUNT = _minStartAmount;
+        UNDELEGATION_PERIOD = _undelegationPeriod;
+    }
+
+    /**
+     * @notice Returns the max amount that a KRO delegator can undelegate.
+     *
+     * @param validator Address of validator.
+     * @param delegator Address of KRO delegator.
+     *
+     * @return The max amount that a KRO delegator can undelegate.
+     */
+    function getKroTotalBalance(
+        address validator,
+        address delegator
+    ) external view returns (uint128) {
+        uint128 shares = _vaults[validator].kroDelegators[delegator].shares;
+        return previewUndelegate(validator, shares);
+    }
+
+    /**
+     * @notice Returns the max amount of KRO that a KGH delegator can undelegate.
+     *
+     * @param validator Address of validator.
+     * @param delegator Address of KGH delegator.
+     * @param tokenId   Token Id of the KGH.
+     *
+     * @return The max amount that a KGH delegator can undelegate.
+     */
+    function getKghTotalBalance(
+        address validator,
+        address delegator,
+        uint256 tokenId
+    ) external view returns (uint128) {
+        KghDelegator storage kghDelegator = _vaults[validator].kghDelegators[delegator];
+        uint128 kghShares = kghDelegator.kghShares[tokenId];
+        uint128 kroShares = kghDelegator.kroShares[tokenId];
+        uint128 kroInKgh = KGH_MANAGER.totalKroInKgh(tokenId);
+
+        uint128 kghAssets = previewKghUndelegate(validator, kghShares) - VKRO_PER_KGH;
+        uint128 kroAssets = previewUndelegate(validator, kroShares) - kroInKgh;
+        return kghAssets + kroAssets;
+    }
+
+    /**
+     * @notice Returns the amount of KRO shares that the KRO delegator has.
+     *
+     * @param validator Address of the validator.
+     * @param delegator Address of the delegator.
+     *
+     * @return The amount of KRO shares that the KRO delegator has.
+     */
+    function getKroTotalShareBalance(
+        address validator,
+        address delegator
+    ) external view returns (uint128) {
+        return _vaults[validator].kroDelegators[delegator].shares;
+    }
+
+    /**
+     * @notice Returns the amount of KGH shares that the KGH delegator has.
+     *
+     * @param validator Address of the validator.
+     * @param delegator Address of the delegator.
+     * @param tokenId   Token Id of the KGH.
+     *
+     * @return The amount of KRO shares that the KGH delegator has.
+     * @return The amount of KGH shares that the KGH delegator has.
+     */
+    function getKghTotalShareBalance(
+        address validator,
+        address delegator,
+        uint256 tokenId
+    ) external view returns (uint128, uint128) {
+        KghDelegator storage kghDelegator = _vaults[validator].kghDelegators[delegator];
+        return (kghDelegator.kroShares[tokenId], kghDelegator.kghShares[tokenId]);
+    }
+
+    /**
+     * @notice Allows an on-chain or off-chain user to simulate the effects of their KRO delegation at the current block.
+     *
+     * @param validator Address of the validator.
+     * @param assets    The amount of assets to delegate.
+     *
+     * @return The amount of shares that the Vault would exchange for the amount of assets provided.
+     */
+    function previewDelegate(
+        address validator,
+        uint128 assets
+    ) public view virtual returns (uint128) {
+        return _convertToKroShares(validator, assets);
+    }
+
+    /**
+     * @notice Allows an on-chain or off-chain user to simulate the effects of their KRO undelegation
+     *         at the current block.
+     *
+     * @param validator The address of the validator.
+     * @param shares    The amount of shares to undelegate.
+     *
+     * @return The amount of assets that the Vault would exchange for the amount of shares provided.
+     */
+    function previewUndelegate(
+        address validator,
+        uint128 shares
+    ) public view virtual returns (uint128) {
+        return _convertToKroAssets(validator, shares);
+    }
+
+    /**
+     * @notice Allows an on-chain or off-chain user to simulate the effects of their KGH delegation
+     *         at the current block given current on-chain conditions.
+     *
+     * @param validator The address of the validator.
+     *
+     * @return The amount of shares that the Vault would exchange for the amount of assets provided.
+     */
+    function previewKghDelegate(address validator) public view virtual returns (uint128) {
+        return _convertToKghShares(validator, VKRO_PER_KGH);
+    }
+
+    /**
+     * @notice Allows an on-chain or off-chain user to simulate the effects of their KGH undelegation
+     *         at the current block, given current on-chain conditions.
+     *
+     * @param validator The address of the validator.
+     * @param tokenId   The tokenId of KGH to undelegate.
+     *
+     * @return The amount of assets that the Vault would exchange for the KGH of given tokenId.
+     */
+    function previewKghUndelegate(
+        address validator,
+        uint256 tokenId
+    ) public view virtual returns (uint128) {
+        return
+            _convertToKghAssets(
+                validator,
+                _vaults[validator].kghDelegators[msg.sender].kghShares[tokenId]
+            );
+    }
+
+    /**
+     * @notice Returns the total amount of KRO assets held by the vault.
+     *
+     * @param validator Address of the validator.
+     *
+     * @return The total amount of KRO assets held by the vault.
+     */
+    function totalKroAssets(address validator) public view virtual returns (uint128) {
+        return _vaults[validator].asset.totalKro;
+    }
+
+    /**
+     * @notice Returns the total number of KGHs held by the vault.
+     *
+     * @param validator Address of the validator.
+     *
+     * @return The total number of KGHs held by the vault.
+     */
+    function totalKghNum(address validator) public view virtual returns (uint128) {
+        return _vaults[validator].asset.totalKgh;
+    }
+
+    /**
+     * @notice Returns the total amount of KGH assets held by the vault.
+     *
+     * @param validator Address of the validator.
+     *
+     * @return The total amount of KGH assets held by the vault.
+     */
+    function _totalKghAssets(address validator) internal view virtual returns (uint128) {
+        Vault storage vault = _vaults[validator];
+        return vault.asset.totalKgh * VKRO_PER_KGH + vault.reward.boostedReward;
+    }
+
+    /**
+     * @notice Returns the total amount of KRO shares held by the vault.
+     *
+     * @param validator Address of the validator.
+     *
+     * @return The total amount of shares held by the validator vault.
+     */
+    function _totalKroShares(address validator) internal view virtual returns (uint128) {
+        return _vaults[validator].asset.totalKroShares;
+    }
+
+    /**
+     * @notice Returns the total amount of KGH shares held by the vault.
+     *
+     * @param validator Address of the validator.
+     *
+     * @return The total amount of shares held by the validator vault.
+     */
+    function _totalKghShares(address validator) internal view virtual returns (uint128) {
+        return _vaults[validator].asset.totalKghShares;
+    }
+
+    /**
+     * @notice Delegate KRO to the validator and returns the amount of shares that the vault would
+     *         exchange.
+     *
+     * @param validator Address of the validator.
+     * @param assets    The amount of KRO to delegate.
+     *
+     * @return The amount of shares that the Vault would exchange for the amount of assets provided.
+     */
+    function delegate(
+        address validator,
+        uint128 assets
+    ) external checkIsActive(validator) returns (uint128) {
+        uint128 shares = previewDelegate(validator, assets);
+        _delegate(validator, msg.sender, assets, shares);
+        emit KroDelegated(validator, msg.sender, assets, shares);
+        return shares;
+    }
+
+    /**
+     * @notice Delegate KGH to the validator and returns the amount of shares that the vault would
+     *         exchange.
+     *
+     * @param validator Address of the validator.
+     * @param tokenId   The tokenId of KGH to delegate.
+     *
+     * @return The amount of KRO shares that the Vault would exchange for the KGH provided.
+     * @return The amount of KGH shares that the Vault would exchange for the KGH provided.
+     */
+    function delegateKgh(
+        address validator,
+        uint256 tokenId
+    ) external checkIsActive(validator) returns (uint128, uint128) {
+        uint128 kroInKgh = KGH_MANAGER.totalKroInKgh(tokenId);
+        uint128 kroShares = previewDelegate(validator, kroInKgh);
+        uint128 kghShares = previewKghDelegate(validator);
+        _delegateKgh(validator, tokenId, kroInKgh, kroShares, kghShares);
+
+        emit KghDelegated(validator, msg.sender, tokenId, kroInKgh, kroShares, kghShares);
+        return (kroShares, kghShares);
+    }
+
+    /**
+     * @notice Delegate KGHs to the validator and returns the amount of shares that the vault would
+     *         exchange.
+     *
+     * @param validator Address of the validator.
+     * @param tokenIds  The token ids of KGHs to delegate.
+     *
+     * @return The amount of KRO shares that the Vault would exchange for the KGHs provided.
+     * @return The amount of KGH shares that the Vault would exchange for the KGHs provided.
+     */
+    function delegateKghBatch(
+        address validator,
+        uint256[] calldata tokenIds
+    ) external checkIsActive(validator) returns (uint128, uint128) {
+        uint128 kroShares;
+        uint128 kghShares = previewKghDelegate(validator);
+        uint128 kroInKghs;
+
+        for (uint256 i = 0; i < tokenIds.length; ) {
+            KGH.safeTransferFrom(msg.sender, address(this), tokenIds[i]);
+
+            uint128 kroInKgh = KGH_MANAGER.totalKroInKgh(tokenIds[i]);
+            uint128 kroSharesForTokenId = previewDelegate(validator, kroInKgh);
+
+            unchecked {
+                kroInKghs += kroInKgh;
+                kroShares += kroSharesForTokenId;
+
+                _vaults[validator].kghDelegators[msg.sender].kroShares[
+                    tokenIds[i]
+                ] = kroSharesForTokenId;
+                _vaults[validator].kghDelegators[msg.sender].kghShares[tokenIds[i]] = kghShares;
+
+                ++i;
+            }
+        }
+
+        kghShares *= uint128(tokenIds.length);
+        _delegateKghBatch(validator, uint128(tokenIds.length), kroInKghs, kroShares, kghShares);
+
+        emit KghBatchDelegated(validator, msg.sender, tokenIds, kroShares, kghShares);
+        return (kroShares, kghShares);
+    }
+
+    /**
+     * @notice Initiate KRO undelegation for given validator and per-mille.
+     *
+     * @param validator Address of the validator.
+     * @param shares    The amount of shares to undelegate.
+     */
+    function initUndelegate(address validator, uint128 shares) external {
+        require(
+            shares > 0 && shares <= _vaults[validator].kroDelegators[msg.sender].shares,
+            "AssetManager: Invalid amount of shares to undelegate"
+        );
+
+        uint128 assets = previewUndelegate(validator, shares);
+        _initUndelegate(validator, msg.sender, assets, shares);
+        emit KroUndelegationInitiated(validator, msg.sender, assets, shares);
+    }
+
+    /**
+     * @notice Initiate KGH undelegation for given validator and tokenId.
+     *
+     * @param validator Address of the validator.
+     * @param tokenId   Token id of KGH to undelegate.
+     */
+    function initUndelegateKgh(address validator, uint256 tokenId) external {
+        uint128 kroShares = _vaults[validator].kghDelegators[msg.sender].kroShares[tokenId];
+        uint128 kghShares = _vaults[validator].kghDelegators[msg.sender].kghShares[tokenId];
+
+        require(kghShares != 0, "AssetManager: No shares for the given tokenId");
+        _initUndelegateKgh(validator, msg.sender, tokenId, kroShares, kghShares);
+
+        emit KghUndelegationInitiated(validator, msg.sender, tokenId, kroShares, kghShares);
+    }
+
+    /**
+     * @notice Initiate KGH undelegation for given validator and token ids.
+     *
+     * @param validator Address of the validator.
+     * @param tokenIds  Array of token ids of KGHs to undelegate.
+     */
+    function initUndelegateKghBatch(address validator, uint256[] calldata tokenIds) external {
+        KghDelegator storage kghDelegator = _vaults[validator].kghDelegators[msg.sender];
+        uint128 kroShares;
+        uint128 kghShares;
+        uint128 kroInKghs;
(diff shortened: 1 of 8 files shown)


packages/contracts/contracts/test/AssetManager.t.sol
```diff
@@ -0,0 +1,608 @@
...
+
+    function replaceOutput(uint256 _outputIndex, bytes32 _outputRoot) external {
+        l2Outputs[_outputIndex] = Types.CheckpointOutput({
+            submitter: msg.sender,
+            outputRoot: _outputRoot,
+            timestamp: uint128(block.timestamp),
+            l2BlockNumber: l2Outputs[_outputIndex].l2BlockNumber
+        });
+    }
+}
```
Pangssu, line 44: What about importing MockL2OutputOracle from another test?

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1466 @@
...
+        uint128 reward = vault[msg.sender].reward.validatorRewardKro;
+        require(reward > 0, "AssetManager: No reward to withdraw");
+
+        vault[msg.sender].reward.validatorRewardKro = 0;
+        bool success = IERC20(GOVERNANCE_TOKEN).transfer(msg.sender, reward);
+        require(success, "AssetManager: Transfer failed");
+    }
+
+    /**
+     * @notice Internal function to add base and boosted reward to the vaults of finalized output submitters.
```
seolaoh, line 895: Please add @return description.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1466 @@
...
+    /**
+     * @notice Modifier to check if the vault is initiated.
+     */
+    modifier checkIsActive(address validator) {
+        require(
+            msg.sender == validator || vault[validator].asset.validatorKro > MIN_REGISTER_AMOUNT,
+            "AssetManager: Vault is inactive"
+        );
+        _;
+    }
```
seolaoh, line 1465: Please move these modifiers above constructor.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1466 @@
...
+            uint32(
+                (vault[validator].asset.totalKro +
+                    vault[validator].reward.boostedReward +
+                    vault[validator].reward.validatorRewardKro) / 1e16
+            )
+        );
+    }
+
+    /**
+     * @dev See {IERC721Receiver-onERC721Received}
```
seolaoh, line 1434:
```suggestion
     * @inheritdoc IERC721Receiver
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1466 @@
...
+    }
+
+    /**
+     * @notice Add pending assets and shares when undelegating KRO.
+     *
+     * @param _vault Vault of the validator.
+     * @param assets The amount of assets to add as pending asset.
+     * @param shares The amount of shares to add as pending share.
+     */
+    function _addPendingKroShares(Vault storage _vault, uint128 assets, uint128 shares) internal {
```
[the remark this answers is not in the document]
sm-stack: The reason why I named it `_vault` was to separate from the mapping `vault`. The naming convention is typically applied to the internal functions and state variables, not to the parameters.

But according to the convention, it would be better to add underscore prefix to the following internal state variables:
- `weightTree`
- `nextRewardOutputIndex`
- `vault`
- `submittedOutputs`
- `pendingChallengeReward`

and then it would be better to change `Vault storage _vault` to `Vault storage vault` in this file. How do you think about this?

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1466 @@
...
+     *
+     * @return The max amount that a KGH delegator can undelegate.
+     */
+    function getMaxUndelegateKghAmount(
+        address validator,
+        address delegator,
+        uint256 tokenId
+    ) external view returns (uint128) {
+        uint128 kghShares = vault[validator].kghDelegators[delegator].kghShares[tokenId];
+        uint128 kroShares = vault[validator].kghDelegators[delegator].kroShares;
```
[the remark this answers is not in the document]
sm-stack: Just tried this on Remix, and found out that it reduces gas by 100 (may differ by context). Thanks for pointing this out!

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1466 @@
...
+        uint128 kroInKgh = uint128(IKGHManager(KGH_MANAGER).totalKroInKgh(tokenId));
+
+        uint128 kghAssets = convertToKghAssets(validator, kghShares) - VKRO_PER_KGH;
+        uint128 kroAssets = convertToKroAssets(validator, kroShares) - kroInKgh;
+        return kghAssets + kroAssets;
+    }
+
+    /**
+     * @notice Returns the maximum amount of the underlying asset that can be delegated into the vault of given validator.
+     * @return The amount of assets that the Vault would exchange for the amount of shares provided.
```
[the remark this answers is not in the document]
sm-stack: That was to follow the standard implementation of ERC-4626, but I agree that it's meaningless. I'll replace it to the constant variable.

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
seolaoh: Oh I see it, that sounds good.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1466 @@
...
+
+    /**
+     * @notice Internal function to undelegate KRO from the validator.
+     *
+     * @param validator Address of the validator.
+     * @param owner     Address of the delegator.
+     * @param assets    The amount of KRO to undelegate.
+     * @param shares    The amount of shares to undelegate.
+     */
+    function _initUndelegate(
```
seolaoh, line 1188: I think if validator self-undelegates and leaves less than `MIN_REGISTER_AMOUNT`, he should remove himself from the weight tree. So in this function, the logic to check it and remove from tree should be added. wdyt?

seolaoh: or more strictly, if less than `MIN_START_AMOUNT`?

packages/contracts/contracts/test/AssetManager.t.sol
sm-stack: That would be better, I'll change it. Thanks!

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1470 @@
...
+     *
+     * @param validator Address of the validator.
+     */
+    function _updateWeightTree(address validator) internal {
+        _weightTree.update(
+            validator,
+            uint32(
+                (_vault[validator].asset.totalKro +
+                    _vault[validator].reward.boostedReward +
+                    _vault[validator].reward.validatorRewardKro) / 1e16
```
seolaoh, line 1454: Are there any reason to divide with `1e16`?

sm-stack: As my first thought, dividing in`1e17` may result in a problem that it cannot reflect the relatively small amount of delegations, and the case you mentioned is not very likely to happen.

But I think it would be better to lower this to `1e13` and increase the bits of parameters in `Node` struct at `BalancedWeightTree` like this:
`struct Node {
        address addr;
        uint40 weight;
        uint40 weightSum;
        uint40 parent;
        uint40 leftChild;
        uint32 rightChild;
        bool isLeftChild;
    }
`
It still occupies 2 slots, so I think the gas will not increase much. How do you think about this? If reasonable, I can update the weight tree with the fix.

seolaoh: That's a good idea. Then how about like below and just remove the denominator?

```solidity
    struct Node {
        address addr;
        uint32 parent;
        uint32 leftChild;
        uint32 rightChild;
        bool isLeftChild;
        uint120 weight;
        uint120 weightSum;
    }
```

sm-stack: That looks much better, I'll fix it.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1466 @@
...
+     *
+     * @custom:field totalKro                   Total amount of KRO in the vault
+     * @custom:field totalKroShares             Total shares for KRO delegation in the vault
+     * @custom:field totalKgh                   Total number of KGH in the vault
+     * @custom:field totalKroInKgh              Total amount of KRO which KGHs in the vault have
+     * @custom:field totalKghShares             Total shares for KGH delegation in the vault
+     * @custom:field validatorKro               Amount of KRO that the validator self-delegated
+     * @custom:field totalPendingAssets         Total pending KRO for undelegation
+     * @custom:field totalPendingBoostedRewards Total pending boosted rewards in KRO for undelegation
+     * @custom:field totalPendingKroShares      Total pending KRO shares for undelegation
```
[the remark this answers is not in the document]
Pangssu: my mistake. `@custom:field` is right.

Pangssu (approved):

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1470 @@
...
+
+    /**
+     * @notice Initiate KRO undelegation for given validator and per-mille.
+     *
+     * @param validator Address of the validator.
+     * @param perMille  The Per-mille of shares to undelegate.
+     */
+    function initUndelegate(address validator, uint128 perMille) external {
+        require(
+            perMille > 0 && perMille <= 1000,
```
seolaoh, line 730:
```suggestion
            perMille > 0 && perMille <= PER_MILLE_DENOM,
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1470 @@
...
+
+    /**
+     * @notice The maximum number of KROs that can be undelegated at once.
+     */
+    uint256 public constant MAX_DELEGATE = 50_000_000 * 10 ** 18;
+
+    /**
+     * @notice Address of the KRO token contract.
+     */
+    IERC20 public immutable GOVERNANCE_TOKEN;
```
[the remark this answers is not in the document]
sm-stack: Yep I'll change it with other fixes.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1470 @@
...
+     * @param rounding  The rounding direction.
+     */
+    function _convertToKroShares(
+        address validator,
+        uint128 assets,
+        Uint128Math.Rounding rounding
+    ) internal view virtual returns (uint128) {
+        return
+            assets.mulDiv(
+                totalKroShares(validator) + 10 ** 6,
```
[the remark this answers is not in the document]
sm-stack: That will be better, but just making `6` as a constant generates compile error due to the casting mechanism of Solidity. I'll make `10 ** 6` as a constant, how do you think?

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1470 @@
...
+     */
+    function _initUndelegateKgh(
+        address validator,
+        address owner,
+        uint256 tokenId,
+        uint128 kroShares,
+        uint128 kghShares
+    ) internal virtual {
+        Vault storage vault = _vault[validator];
+        uint128 kroAssetsToWithdraw = previewUndelegate(validator, kroShares);
```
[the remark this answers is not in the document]
sm-stack: You're right, I missed that. I'll fix it, thanks for pointing it out!

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1470 @@
...
+        uint128 kroInKgh = uint128(KGH_MANAGER.totalKroInKgh(tokenId));
+        return _convertToKghShares(validator, kroInKgh, Uint128Math.Rounding.Down);
+    }
+
+    function convertToKghAssets(
+        address validator,
+        uint128 shares
+    ) public view virtual returns (uint128) {
+        return _convertToKghAssets(validator, shares, Uint128Math.Rounding.Down);
+    }
```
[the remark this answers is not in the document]
sm-stack: That's right, and I'll delete the `convertTo**` cases. Also there's no cases for `Uint128Math.Rounding.Up`, I'll fix the Uint128Math library and corresponding codes.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1470 @@
...
+    /**
+     * @notice Allows an on-chain or off-chain user to simulate the effects of their KGH delegation
+     *         at the current block given current on-chain conditions.
+     *
+     * @param validator The address of the validator.
+     * @param assets    The amount of assets to delegate.
+     *
+     * @return The amount of shares that the Vault would exchange for the amount of assets provided.
+     */
+    function previewKghDelegate(
```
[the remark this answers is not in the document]
sm-stack: That would be better, thanks!

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1470 @@
...
+     * @return The amount of KRO shares that the Vault would exchange for the KGH provided.
+     * @return The amount of KGH shares that the Vault would exchange for the KGH provided.
+     */
+    function delegateKgh(
+        address validator,
+        uint256 tokenId
+    ) external checkIsActive(validator) returns (uint128, uint128) {
+        uint128 kroInKgh = uint128(KGH_MANAGER.totalKroInKgh(tokenId));
+        uint128 kroShares = previewDelegate(validator, kroInKgh);
+        uint128 kghShares = previewKghDelegate(validator, VKRO_PER_KGH);
```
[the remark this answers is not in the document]
sm-stack: That was how ERC-4626 is implemented, I think that's for making the code more readable.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1470 @@
...
+    function finalizeUndelegateKgh(address validator) external returns (uint128) {
+        KghDelegator storage kghDelegator = _vault[validator].kghDelegators[msg.sender];
+        uint256[] memory requestTimes = kghDelegator.undelegateRequestTimes;
+        uint128 assets = finalizeUndelegate(validator);
+        uint128 shares;
+
+        for (uint256 i = 0; i < requestTimes.length; ) {
+            if (requestTimes[i] + WITHDRAWAL_DELAY <= block.timestamp) {
+                uint128 kghSharesToBurn = kghDelegator.undelegateRequests[requestTimes[i]];
+                uint256[] memory tokenId = kghDelegator.undelegateRequestedIds[requestTimes[i]];
```
seolaoh, line 830:
```suggestion
                uint256[] memory tokenIds = kghDelegator.undelegateRequestedIds[requestTimes[i]];
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1428 @@
...
+    /**
+     * @notice Finalize all pending KGH undelegation and returns the amount of assets that the vault would
+     *         exchange for the amount of shares provided.
+     *
+     * @param validator Address of the validator.
+     *
+     * @return The amount of assets that the vault would exchange for the amount of shares provided.
+     */
+    function finalizeUndelegateKgh(address validator) external returns (uint128) {
+        KghDelegator storage kghDelegator = _vault[validator].kghDelegators[msg.sender];
```
seolaoh, line 822: Isn't it better to check below instead of L832?
```suggestion
        require(_vault[validator].asset.totalPendingKghShares > 0, "");

        KghDelegator storage kghDelegator = _vault[validator].kghDelegators[msg.sender];
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1470 @@
...
+        uint256[] memory requestTimes = kghDelegator.undelegateRequestTimes;
+        uint128 assets = finalizeUndelegate(validator);
```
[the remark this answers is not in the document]
seolaoh: Also if we remain as is, this line should be like below because all assets related to base reward is already handled in `finalizeUndelegate`.
```suggestion
        finalizeUndelegate(validator);
        uint128 assets;
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1428 @@
...
+                    }
+                }
+
+                delete kghDelegator.undelegateRequests[requestTimes[i]];
+                delete kghDelegator.undelegateRequestTimes[i];
+            }
+
+            unchecked {
+                ++i;
+            }
```
seolaoh, line 856: If the `undelegateRequestTimes` is pushed orderly, we can break the for loop earlier.
```suggestion
                unchecked {
                    ++i;
                }
            } else {
                break;
            }
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1428 @@
...
+
+                for (uint256 tokenIdIndex = 0; tokenIdIndex < tokenId.length; ) {
+                    KGH.safeTransferFrom(address(this), msg.sender, tokenId[tokenIdIndex]);
+
+                    unchecked {
+                        ++tokenIdIndex;
+                    }
+                }
+
+                delete kghDelegator.undelegateRequests[requestTimes[i]];
```
seolaoh, line 850:
```suggestion
                delete kghDelegator.undelegateRequests[requestTimes[i]];
                delete kghDelegator.undelegateRequestedIds[requestTimes[i]];
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1428 @@
...
+                delete kghDelegator.undelegateRequestTimes[i];
+            }
+
+            unchecked {
+                ++i;
+            }
+        }
+
+        _vault[validator].asset.totalPendingAssets -= assets;
```
seolaoh, line 859:
```suggestion
        _vault[validator].asset.totalPendingBoostedRewards -= assets;
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1428 @@
...
+        _vault[validator].asset.totalPendingKghShares -= shares;
+
+        bool success = IERC20(TOKEN).transfer(msg.sender, assets);
```
seolaoh, line 862: Are there any reason not to use `SafeERC20.SafeTransfer`?

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
seolaoh: looks good to me!

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1428 @@
...
+     * @notice Constructs the delegator of KRO in the vault of a validator.
+     *
+     * @custom:field shares                 Amount of shares for KRO delegation.
+     * @custom:field undelegateRequestTimes Timestamps of undelegation requests.
+     * @custom:field undelegateRequests     Pending shares at timestamps undelegations are requested.
+     */
+    struct KroDelegator {
+        uint128 shares;
+        uint256[] undelegateRequestTimes;
+        mapping(uint256 => uint128) undelegateRequests;
```
0xHansLee, line 82: Why don't we rename to `pendingShares`? The term `Request` is quite ambiguous to me.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1428 @@
...
+    }
+
+    /**
+     * @notice Constructs the vault of a validator.
+     *
+     * @custom:field asset             Asset information of the vault.
+     * @custom:field reward            Reward information of the vault.
+     * @custom:field isInitiated       Whether the vault is initiated.
+     * @custom:field noSubmissionCount Number of counts that the validator did not submit the output in priority round.
```
0xHansLee, line 110: `kroDelegators` and `kghDelegators` are missing

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1428 @@
...
+     */
+    struct Vault {
+        Asset asset;
+        Reward reward;
+        bool isInitiated;
```
0xHansLee, line 115: This field is not used anywhere in `AssetManager` contract, but it will be used in `ValidatorManager`?
And why don't rename to `isActivated`?

packages/contracts/contracts/L1/AssetManager.sol
sm-stack: But if then, there could be an edge case that delegators withdraw all their tokens from the vault, so the validator vault has low amount of tokens but still that validator remains at the tree. How do you think about this?

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1470 @@
...
+    function _updateWeightTree(address validator) internal {
+        _weightTree.update(
+            validator,
+            uint120(
+                (_vault[validator].asset.totalKro +
+                    _vault[validator].reward.boostedReward +
+                    _vault[validator].reward.validatorRewardKro)
+            )
+        );
+    }
```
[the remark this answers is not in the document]
sm-stack: That would be better, I'll fix it. Thanks!

packages/contracts/contracts/L1/AssetManager.sol
sm-stack: That would be much better, thanks!

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1428 @@
...
+        for (; rewardedNum < MAX_REWARD_RECIPIENTS; ) {
+            uint256 expiryTime = _submittedOutputs[outputIndex];
+            if (block.timestamp >= expiryTime && expiryTime != 0) {
+                delete _submittedOutputs[outputIndex];
+                output = L2_ORACLE.getL2Output(outputIndex);
+                _increaseBalanceWithReward(output.submitter);
+
+                uint128 challengeReward = _pendingChallengeReward[outputIndex];
+                if (challengeReward > 0) {
+                    _modifyBalanceWithSlashing(_vault[output.submitter], challengeReward, false);
```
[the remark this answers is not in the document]
sm-stack: That's a good idea, I'll add them.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1428 @@
...
+                    delete _pendingChallengeReward[outputIndex];
+                }
+
```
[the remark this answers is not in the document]
sm-stack: That's what I missed, thanks!

packages/contracts/contracts/L1/AssetManager.sol
sm-stack: Yep it will be used only in `ValidatorManager`. How do you think about this suggestion @seolaoh ?

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1428 @@
...
+     * @custom:field commissionRateNumerator       Numerator of the commission rate.
+     * @custom:field commissionRateMaxChangeRange  Range that a commission rate can be changed to.
+     * @custom:field lastCommissionRateChangedTime Last timestamp when the commission rate was changed.
+     */
+    struct Reward {
+        uint128 boostedReward;
+        uint128 validatorRewardKro;
+        uint8 commissionRateNumerator;
+        uint8 commissionRateMaxChangeRange;
+        uint128 lastCommissionRateChangedTime;
```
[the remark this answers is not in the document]
sm-stack: That looks more simple. Thanks!

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1428 @@
...
+        require(success, "AssetManager: Transfer failed");
+
+        emit KghUndelegationFinalized(validator, msg.sender, assets);
+        return assets;
+    }
+
+    /**
+     * @notice Withdraw the reward of the validator.
+     */
+    function withdrawValidatorReward() external {
```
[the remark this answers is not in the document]
sm-stack: That's definitely right, I'll add the `initWithdraw` and `finalizeWithdrawal` functions for validators, and the corresponding tests for them.

packages/contracts/contracts/L1/AssetManager.sol
seolaoh: Yes it will be used in `ValidatorManager` and validator can be initiated but in `inactive` status, so the name is appropriate to separate the status. Let's check it in the ValidatorManager PR.

packages/contracts/contracts/L1/AssetManager.sol
sm-stack: I agree with that, could you check the following namings are sufficient?

- `undelegateRequests` (at `KroDelegator`) -> `pendingKroShares` 
- `undelegateRequests` (at `KghDelegator`) -> `pendingKghShares`
- `undelegateRequestedIds` -> `pendingTokenIds`

packages/contracts/contracts/L1/AssetManager.sol
seolaoh: Yes, such edge cases can be problematic, but I'm worried that undelegation will be too costly for delegators of small amount. It also seems to happen fairly often that the total KRO goes below `MIN_START_AMOUNT` by an undelegation, but it doesn't seem to happen as often that all delegators leave but validators keep their KRO above `MIN_REGISTER_AMOUNT`.

packages/contracts/contracts/L1/AssetManager.sol
sm-stack: Found an edge case of this: An user cannot withdraw KGH from the vault if `boostedReward` is 0. `_addPendingKghShares` will not increase `_vault[validator].asset.totalPendingKghShares`, since the `boostedReward` and corresponding parameter `kghShares` are 0. How do you think about this?

packages/contracts/contracts/L1/AssetManager.sol
0xHansLee: All looks good. But how about `pendingKghIds`?

BTW, for abbreviation, capitals are used in golang, but not in contract. If so, let's use `kgh`.
(e.g. kgh -> KGH, Id -> ID)

packages/contracts/contracts/L1/AssetManager.sol
0xHansLee: OK!

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+    struct KroDelegator {
+        uint128 shares;
+        uint256[] undelegateRequestTimes;
+        mapping(uint256 => uint128) undelegateRequests;
+    }
+
+    /**
+     * @notice Constructs the delegator of KGH in the vault of a validator.
+     *
+     * @custom:field kroShares              Amount of shares for KRO delegation.
```
seolaoh, line 89:
```suggestion
     * @custom:field kroShares              Amount of shares for KRO in KGH with a corresponding tokenId.
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+     */
+    uint256 internal _nextRewardOutputIndex;
+
+    /**
+     * @notice A mapping of validator address to the vault.
+     */
+    mapping(address => Vault) internal _vault;
+
+    /**
+     * @notice A mapping of the submitted output index to submission timestamp.
```
seolaoh, line 240:
```suggestion
     * @notice A mapping of the submitted output index to finalization timestamp.
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+     */
+    mapping(uint256 => uint256) internal _submittedOutputs;
+
+    /**
+     * @notice A mapping of challenge timestamp to pending challenge rewards.
```
seolaoh, line 245:
```suggestion
     * @notice A mapping of challenged output index to pending challenge rewards.
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+
+    /**
+     * @notice The address of the SecurityCouncil contract. Can be updated via upgrade.
+     */
+    address public immutable SECURITY_COUNCIL;
+
+    /**
+     * @notice The max number of reward recipients when distributing rewards.
+     */
+    uint128 public immutable MAX_REWARD_RECIPIENTS;
```
0xHansLee, line 192: This may seem like it means the number of delegates receiving rewards. How about renaming `MAX_OUTPUT_FINALIZATION`?

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+
+    /**
+     * @notice Minimum amount to register as a validator.
+     */
+    uint128 public immutable MIN_REGISTER_AMOUNT;
+
+    /**
+     * @notice Minimum amount to start submitting outputs.
+     */
+    uint128 public immutable MIN_START_AMOUNT;
```
0xHansLee, line 217: `ACTIVATION_THRESHOLD`?

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+
+    /**
+     * @notice Withdrawal delay for the undelegation.
+     */
+    uint256 public immutable WITHDRAWAL_DELAY;
```
0xHansLee, line 222: `PENDING_PERIOD` or `UNDELEGATION_PERIOD`?

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+
+    /**
+     * @notice Weighted tree to store and calculate the probability to be selected as an output submitter.
+     */
+    BalancedWeightTree.Tree internal _weightTree;
```
0xHansLee, line 227: The weightTree is a data structure, so I recommend to use the name for this `_validatorTree`.
```suggestion
    BalancedWeightTree.Tree internal _validatorTree;
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+            kghShares += kghShareForTokenId;
+            kroInKghs += KGH_MANAGER.totalKroInKgh(tokenIds[i]);
+            kghAssetsToWithdraw += previewKghUndelegate(validator, tokenIds[i]);
+
+            delete _vault[validator].kghDelegators[msg.sender].kroShares[tokenIds[i]];
+            delete _vault[validator].kghDelegators[msg.sender].kghShares[tokenIds[i]];
+
+            unchecked {
+                ++i;
+            }
```
seolaoh, line 768:
```suggestion

            unchecked {
                kroShares += kroSharesForTokenId;
                kghShares += kghShareForTokenId;
                kroInKghs += KGH_MANAGER.totalKroInKgh(tokenIds[i]);
                kghAssetsToWithdraw += previewKghUndelegate(validator, tokenIds[i]);

                delete _vault[validator].kghDelegators[msg.sender].kroShares[tokenIds[i]];
                delete _vault[validator].kghDelegators[msg.sender].kghShares[tokenIds[i]];

                ++i;
            }
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+
+    /**
+     * @notice The output index to distribute reward next.
+     */
+    uint256 internal _nextRewardOutputIndex;
+
+    /**
+     * @notice A mapping of validator address to the vault.
+     */
+    mapping(address => Vault) internal _vault;
```
0xHansLee, line 237:
```suggestion
    mapping(address => Vault) internal _vaults;
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+        unchecked {
+            vault.asset.totalPendingBoostedRewards += boostedRewards;
+            vault.asset.totalPendingKghShares += kghShares;
+            vault.kghDelegators[msg.sender].undelegateRequests[block.timestamp] += kghShares;
+            vault.kghDelegators[msg.sender].undelegateRequestTimes.push(block.timestamp);
+        }
+    }
+
+    /**
+     * @notice Internal conversion function for KRO (from assets to shares) with support for rounding direction.
```
seolaoh, line 1014: Same all below.
```suggestion
     * @notice Internal conversion function for KRO (from assets to shares).
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+    ) internal view virtual returns (uint128) {
+        return
+            shares.mulDiv(
+                totalKroAssets(validator) + 1,
+                totalKroShares(validator) + DECIMAL_OFFSET
+            );
+    }
+
+    /**
+     * @notice Internal conversion function for KRO in KGH (from assets to shares) with support for
```
seolaoh, line 1049: It's not for KRO in KGH, for `VKRO_PER_KGH`.
```suggestion
     * @notice Internal conversion function for KGH (from assets to shares) with support for
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+
+    /**
+     * @notice A mapping of validator address to the vault.
+     */
+    mapping(address => Vault) internal _vault;
+
+    /**
+     * @notice A mapping of the submitted output index to submission timestamp.
+     */
+    mapping(uint256 => uint256) internal _submittedOutputs;
```
0xHansLee, line 242: This can be queried from L2 output oracle contract. Do you have any specific reason for this?

packages/contracts/contracts/L1/AssetManager.sol
0xHansLee, line 245:
```suggestion
     * @notice A mapping of output index challenged successfully to pending challenge rewards.
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+     */
+    event ChallengeRewardDistributed(address recipient, uint128 amount);
+
+    /**
+     * @notice Emitted when the output reward is distributed.
+     *
+     * @param recipient Address of the reward recipient.
+     * @param amount    The amount of reward.
+     */
+    event RewardDistributed(address recipient, uint128 amount);
```
0xHansLee, line 389: Why don't we indexing the recipient address?

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+        );
+        _;
+    }
+
+    /**
+     * @notice Modifier to check if the vault is active.
+     */
+    modifier checkIsActive(address validator) {
+        require(
+            msg.sender == validator || _vault[validator].asset.validatorKro >= MIN_REGISTER_AMOUNT,
```
0xHansLee, line 416: The `msg.sender` should be validator here?

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+     * @param _securityCouncil        Address of the SecurityCouncil contract.
+     * @param _token                  Address of the KRO token.
+     * @param _kgh                    Address of the KGH token.
+     * @param _kghManager             Address of the KGHManager contract.
+     * @param _maxRewardRecipients    Max number of reward recipients.
+     * @param _baseReward             Base reward for the validator.
+     * @param _slashingRateNumerator  Numerator of the slashing rate.
+     * @param _minSlashingAmount      Minimum amount to slash.
+     * @param _minRegisterAmount      Minimum amount to register as a validator.
+     * @param _withdrawalDelay        Withdrawal delay for the undelegation.
```
0xHansLee, line 435: Let's make the same order with args of constructor.
And `_minStartAmount` is missing.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+                delete delegator.undelegateRequestTimes[i];
+            }
+
+            unchecked {
+                ++i;
+            }
+        }
+
+        _vault[validator].asset.totalPendingAssets -= assets;
+        _vault[validator].asset.totalPendingKroShares -= shares;
```
seolaoh, line 823: How about this? Also in `finalizeUndelegateKgh`.
```suggestion
        unchecked {
            _vault[validator].asset.totalPendingAssets -= assets;
            _vault[validator].asset.totalPendingKroShares -= shares;
        }
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1470 @@
...
+        return
+            shares.mulDiv(
+                totalKroAssets(validator) + 1,
+                totalKroShares(validator) + 10 ** 6,
+                rounding
+            );
+    }
+
+    /**
+     * @notice Internal conversion function for KGH (from assets to shares) with support for
```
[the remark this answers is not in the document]
seolaoh: Oh I'm sorry, I was wrong.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+        Vault storage vault = _vault[validator];
+
+        unchecked {
+            vault.asset.totalKro += kroInKghs;
+            vault.asset.totalKroInKgh += kroInKghs;
+            vault.asset.totalKgh += kghCount;
+
+            vault.asset.totalKroShares += kroShares;
+            vault.asset.totalKghShares += kghShares;
+        }
```
seolaoh, line 1174:
```suggestion
        Asset storage asset = vault.asset;

        unchecked {
            asset.totalKro += kroInKghs;
            asset.totalKroInKgh += kroInKghs;
            asset.totalKgh += kghCount;

            asset.totalKroShares += kroShares;
            asset.totalKghShares += kghShares;
        }
```

packages/contracts/contracts/L1/AssetManager.sol
seolaoh: Yup I undenstand it, thank you!

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+
+    /**
+     * @notice Decimals offset for the KRO and KGH shares.
+     */
+    uint128 public constant DECIMAL_OFFSET = 10 ** 6;
+
+    /**
+     * @notice The maximum number of KROs that can be undelegated at once.
+     */
+    uint256 public constant MAX_DELEGATE = 50_000_000 * 10 ** 18;
```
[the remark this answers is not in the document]
seolaoh: I think there will be too many constructor arguments. Currently, there are too many arguments for `ValidatorManager`, so I needed to separate the arguments into a struct.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+
+    /**
+     * @notice The denominator for per-mille used at slashing and initialization of undelegation.
+     */
+    uint128 public constant PER_MILLE_DENOM = 1000;
+
+    /**
+     * @notice Virtual KRO amount per KGH.
+     */
+    uint128 public constant VKRO_PER_KGH = 100e18;
```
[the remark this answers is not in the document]
seolaoh: This is not for KRO in KGH, it's only for the KGH NFT itself.

packages/contracts/contracts/L1/AssetManager.sol
seolaoh: `start` means the validator is now added to the tree and can be selected as priority validator. So it is different from `active`, let's check it in the next PR.

packages/contracts/contracts/L1/AssetManager.sol
seolaoh: Cause the validator can also delegate self, so in the case where `msg.sender` is the validator itself, we should not check if the `validatorKro` is more than `MIN_REGISTER_AMOUNT`.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+        uint256[] memory requestTimes = kghDelegator.undelegateRequestTimes;
+        uint128 assetsToUndelegate = finalizeUndelegate(validator);
+        uint128 sharesToUndelegate;
+
+        for (uint256 i = 0; i < requestTimes.length; ) {
+            if (requestTimes[i] + WITHDRAWAL_DELAY <= block.timestamp) {
+                uint128 kghSharesToBurn = kghDelegator.undelegateRequests[requestTimes[i]];
+                uint256[] memory tokenId = kghDelegator.undelegateRequestedIds[requestTimes[i]];
+
+                if (asset.totalPendingKghShares > 0) {
```
seolaoh, line 852:
```suggestion
                uint256[] memory tokenIds = kghDelegator.undelegateRequestedIds[requestTimes[i]];

                if (asset.totalPendingKghShares > 0) {
                    uint128 kghSharesToBurn = kghDelegator.undelegateRequests[requestTimes[i]];
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+     * @param kghShares The amount of KGH shares to undelegate.
+     */
+    function _initUndelegateKgh(
+        address validator,
+        address owner,
+        uint256 tokenId,
+        uint128 kroShares,
+        uint128 kghShares
+    ) internal virtual {
+        Vault storage vault = _vault[validator];
```
seolaoh, line 1226: How about using `asset` below?
```suggestion
        Vault storage vault = _vault[validator];
        Asset storage asset = vault.asset;
```

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
sm-stack: I agree with @seolaoh, and also there are some variables that must not be changed like `DECIMAL_OFFSET` or `VKRO_PER_KGH`, which can lead to inconsistent states if changed after the deployment.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+
+    /**
+     * @notice The maximum number of KROs that can be undelegated at once.
+     */
+    uint256 public constant MAX_DELEGATE = 50_000_000 * 10 ** 18;
+
+    /**
+     * @notice Address of the KRO token contract.
+     */
+    IERC20 public immutable TOKEN;
```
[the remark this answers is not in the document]
sm-stack: `TOKEN` was a naming convention for ERC-4626, but I think that would also be good. How do you think @seolaoh?

packages/contracts/contracts/L1/AssetManager.sol
sm-stack: That would be more intuitive, I'll fix it. Thanks!

packages/contracts/contracts/L1/AssetManager.sol
sm-stack: I missed that point, will add one. Thanks!

packages/contracts/contracts/L1/AssetManager.sol
sm-stack: A validator should not get an opportunity to submit output if he has less amount of KRO than `minStartAmount`, not `minRegisterAmount`? I agree with the point that the gas that users spend will increase in some cases, but I think it will be somewhat problematic if unqualified validators are not removed from the tree.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+    /**
+     * @notice Returns the total amount of KGH assets held by the vault.
+     *
+     * @param validator Address of the validator.
+     *
+     * @return The total amount of KGH assets held by the vault.
+     */
+    function totalKghAssets(address validator) public view virtual returns (uint128) {
+        Vault storage vault = _vault[validator];
+        return vault.asset.totalKgh * VKRO_PER_KGH + vault.reward.boostedReward;
```
0xHansLee, line 588: Should `vault.asset.totalKroInKgh` be included in Kgh asset as well?

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
0xHansLee: Ah, I see. Then, the value 100e18 is a temp value?

packages/contracts/contracts/L1/AssetManager.sol
0xHansLee: Ah, validator also delegates using `delegate` function. I see. TY!

packages/contracts/contracts/L1/AssetManager.sol
seolaoh: Yes the validator misses the output submission opportunity when he has less than `minStartAmount`. I agree with your point, then let's just remain it and discuss about it after investigating the gas usage when we do QA.

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
seolaoh: Yes! It's only used to represent virtual KRO for KGH to distribute boosted reward to the KGH delegators more fairly.

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
seolaoh: Yep let's change it to `ASSET_TOKEN`

packages/contracts/contracts/L1/AssetManager.sol
seolaoh: KRO in KGH is handled using `kroShares`, so this function should exclude that value to calculate only boosted reward not base reward.
By the way, I think to change `total**` functions except for `totalKroAssets` to internal or private functions would be better. Because the return values of these functions include `VKRO_PER_KGH` or share offsets, so not appropriate for public purpose. I think `preview**` functions are enough for public use. Instead, how about remaining only `totalKroAssets` and adding `totalKghNum` function to return `_vault[validator].asset.totalKgh`?

sm-stack: I agree to @seolaoh, that kind transition from `public` to `internal` functions will reduce the deployment cost.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+        }
+    }
+
+    /**
+     * @notice Internal conversion function for KRO (from assets to shares) with support for rounding direction.
+     *
+     * @param validator Address of the validator.
+     * @param assets    The amount of assets to convert to shares.
+     */
+    function _convertToKroShares(
```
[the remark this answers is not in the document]
sm-stack: I thought that the functions are 'converting' the given asset amounts to share, but I agree that the `calculate*` one would also be good. But these are the conventions defined at ERC-4626, so how about leaving them to let devs aware of the fact that this vault architecture inherits it in terms of how to manage states?

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+    /**
+     * @notice Allows an on-chain or off-chain user to simulate the effects of their KRO undelegation
+     *         at the current block.
+     *
+     * @param validator The address of the validator.
+     * @param shares    The amount of shares to undelegate.
+     *
+     * @return The amount of assets that the Vault would exchange for the amount of shares provided.
+     */
+    function previewUndelegate(
```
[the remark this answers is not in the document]
sm-stack: This is the convention from ERC-4626, so how about leaving it like this regarding the same point as above?

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
sm-stack: That was a convention of ERC-4626, and it was intended to prevent the cases that `receiver` (deposit to another account is allowed at ERC-4626) takes more than the max amount he can receive.

I fully agree that it is useless in this case, so I'll remove it. Thanks!

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+
+    /**
+     * @notice Returns the max amount that a KRO delegator can undelegate.
+     *
+     * @param validator Address of validator.
+     * @param delegator Address of KRO delegator.
+     *
+     * @return The max amount that a KRO delegator can undelegate.
+     */
+    function getMaxUndelegateKroAmount(
```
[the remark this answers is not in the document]
sm-stack: That's much better, I'll fix them. Thanks!

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
0xHansLee: Oh, I got it! Then, let's keep this way.

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
0xHansLee: Got it!

packages/contracts/contracts/L1/AssetManager.sol
sm-stack: I think I missed this point. I'll fix it, thanks!

sm-stack: I think this can be done in the next PR of `ValidatorManager`, since the public getter for outputs in L2 Output Oracle contract is already implemented in that contract. @seolaoh could you handle this?

packages/contracts/contracts/L1/AssetManager.sol
0xHansLee: Thank you for your detailed explanation.
In other words, it is to distribute the boosted reward to all the KGH holders fairly? independent of the referral code? AFAIK, each KGH has different amount of KRO in KGH, but this is not considered in distribution of boosted reward. Do I understand, right?

BTW, I understood the point @seolaoh said and agreed with adding prefix of `_` for internal usage.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1470 @@
...
+                unchecked {
+                    assets += sharesToBurn.mulDiv(
+                        _vault[validator].asset.totalPendingAssets,
+                        _vault[validator].asset.totalPendingKroShares
+                    );
+                    shares += sharesToBurn;
+                }
+
+                delete delegator.undelegateRequests[requestTimes[i]];
+                delete delegator.undelegateRequestTimes[i];
```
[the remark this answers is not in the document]
sm-stack: That's true, I'll consider some alternatives to resolve this problem.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1470 @@
...
+            unchecked {
+                ++i;
+            }
+        }
+
+        _vault[validator].asset.totalPendingAssets -= assets;
+        _vault[validator].asset.totalPendingKroShares -= shares;
+
+        bool success = IERC20(GOVERNANCE_TOKEN).transfer(msg.sender, assets);
+        require(success, "AssetManager: Transfer failed");
```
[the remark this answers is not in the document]
sm-stack: That looks much better, I'll change it.

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
sm-stack: The first one looks better in terms of gas, so I'll consider applying it. Thanks!

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+        return (kroShares, kghShares);
+    }
+
+    /**
+     * @notice Initiate KRO undelegation for given validator and per-mille.
+     *
+     * @param validator Address of the validator.
+     * @param perMille  The Per-mille of shares to undelegate.
+     */
+    function initUndelegate(address validator, uint128 perMille) external {
```
0xHansLee, line 707: Do you have any specific reason of using ratio for undelegation? 
IMHO, we can simply implement this with argument of the amount of share.

packages/contracts/contracts/L1/AssetManager.sol
seolaoh: Yes boosted reward is distributed uniformly to all KGH holders using `_convertToKghAssets`, and for KRO in KGH, the base reward will be distributed according to the number of KRO in KGH using `_convertToKroAssets`. So `totalKghAssets` should not include `totalKroInKgh`.

What I suggest is like below:
- convert `totalKghAssets`, `totalKroShares` and `totalKghShares` functions to internal or private functions
- add `totalKghNum` to return the total number of KGH delegated to the validator

packages/contracts/contracts/L1/AssetManager.sol
seolaoh: Sure

packages/contracts/contracts/L1/AssetManager.sol
0xHansLee: Good. I like it.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+     * @param boostedRewards The amount of boosted rewards to add as pending asset.
+     * @param kghShares      The amount of KGH shares to add as pending share.
+     */
+    function _addPendingKghShares(
+        Vault storage vault,
+        uint256[] memory tokenIds,
+        uint128 boostedRewards,
+        uint128 kghShares
+    ) internal {
+        for (uint256 i; i < tokenIds.length; ) {
```
0xHansLee, line 995:
```suggestion
        for (uint256 i = 0; i < tokenIds.length; ) {
```

packages/contracts/contracts/L1/AssetManager.sol
sm-stack: That would make the implementation simpler, but I thought about the integration with frontend, and it would be better to have a precentage-based argument rather than the amount of share itself. How do you think about this?

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1459 @@
...
+        address owner,
+        uint256 tokenId,
+        uint128 kroShares,
+        uint128 kghShares
+    ) internal virtual {
+        Vault storage vault = _vault[validator];
+        uint128 kroAssetsToWithdraw = previewUndelegate(validator, kroShares);
+        uint128 kghAssetsToWithdraw = previewKghUndelegate(validator, tokenId);
+
+        uint256[] memory tokenIdArr = new uint256[](1);
```
0xHansLee, line 1230: For consistent name with batch case,
```suggestion
        uint256[] memory tokenIds = new uint256[](1);
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1470 @@
...
+
+        unchecked {
+            vault.asset.totalKroShares -= kroShares;
+            vault.kghDelegators[msg.sender].kroShares -= kroShares;
+            vault.asset.totalKghShares -= kghShares;
+
+            uint128 baseRewardsToReceive = kroAssetsToWithdraw - kroInKghs;
+            uint128 boostedRewardsToReceive = kghAssetsToWithdraw - VKRO_PER_KGH * uint128(tokenIds.length);
+
+            vault.asset.totalKro -= kroInKghs;
```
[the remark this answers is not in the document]
0xHansLee: In this case, it should be `kroAssetsToWithdraw` as done in `_initUndelegateKgh`, right?

packages/contracts/contracts/L1/AssetManager.sol
0xHansLee: In that case, I think it would be better to use the amount of share token itself.
The way we need to prefer should be to make contract as simple as possible.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1461 @@
...
+     *         exchange for the pending KRO shares.
+     *
+     * @param validator Address of the validator.
+     *
+     * @return The amount of assets that the vault would exchange for the pending KRO shares.
+     */
+    function finalizeUndelegate(address validator) public returns (uint128) {
+        if (_vault[validator].asset.totalPendingKroShares == 0) {
+            return 0;
+        }
```
0xHansLee, line 798: If there is no pending shares, why don't we revert this transaction?

packages/contracts/contracts/L1/AssetManager.sol
seolaoh: This change is applied wrong line (not L75, should be L89)

packages/contracts/contracts/L1/AssetManager.sol
sm-stack: There can be an edge case since `finalizeUndelegateKgh` is using `finalizeUndelegate` inside itself. If a KGH delegator undelegates his KGH without receiving any reward, he cannot finalize his undelegation since his `pendingKroShare` is 0. 

But in the above conversation, I'm currently considering about managing the `pendingKroShares` separately for KGH undelegation, so if then it may be possible to add revert statement. Thanks for pointing this out!

packages/contracts/contracts/L1/AssetManager.sol
sm-stack: Ok, then I'll address it. Thanks!

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1468 @@
...
+     * @return The amount of shares that the Vault would exchange for the amount of assets provided.
+     */
+    function delegate(
+        address validator,
+        uint128 assets
+    ) external checkIsActive(validator) returns (uint128) {
+        require(
+            assets <= ASSET_TOKEN.balanceOf(msg.sender),
+            "AssetManager: Exceeds max delegate amount"
+        );
```
seolaoh, line 636: This would be checked in `ERC20.transferFrom` function, so I think it is unnecessary.

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
sm-stack: Yep I also think `kroAssetsToWithdraw` is right in this place.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1468 @@
...
+            asset.totalKroShares -= kroShares;
+            asset.totalKghShares -= kghShares;
+            delete vault.kghDelegators[owner].kroShares[tokenId];
+            delete vault.kghDelegators[owner].kghShares[tokenId];
+
+            uint128 kroInKgh = KGH_MANAGER.totalKroInKgh(tokenId);
+            uint128 baseRewardsToReceive = kroAssetsToWithdraw - kroInKgh;
+            uint128 boostedRewardsToReceive = kghAssetsToWithdraw - VKRO_PER_KGH;
+
+            asset.totalKro -= kroInKgh;
```
seolaoh, line 1252:
```suggestion
            asset.totalKro -= kroAssetsToWithdraw;
```

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
seolaoh: Oh right, I edited the comment.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1468 @@
...
+     * @notice Initiate KGH undelegation for given validator and tokenIds.
+     *
+     * @param validator Address of the validator.
+     * @param tokenIds  Array of token ids of KGHs to undelegate.
+     */
+    function initUndelegateKghBatch(address validator, uint256[] calldata tokenIds) external {
+        uint128 kroShares;
+        uint128 kghShares;
+        uint128 kroInKghs;
+        uint128 kghAssetsToWithdraw;
```
seolaoh, line 760: Isn't it cheaper to use it below?
```suggestion
        uint128 kghAssetsToWithdraw;
        KghDelegator storage kghDelegator = _vault[validator].kghDelegators[msg.sender];
```

packages/contracts/contracts/L1/AssetManager.sol
sm-stack: I'll fix it, thanks :)

packages/contracts/contracts/L1/AssetManager.sol
sm-stack: `UNDELEGATION_PERIOD` looks good, I'll change to that.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1468 @@
...
+        SafeERC20.safeTransfer(ASSET_TOKEN, msg.sender, assetsToUndelegate);
+
+        emit KghUndelegationFinalized(validator, msg.sender, assetsToUndelegate);
+        return assetsToUndelegate;
+    }
+
+    /**
+     * @notice Withdraw the reward of the validator.
+     */
+    function withdrawValidatorReward() external {
```
0xHansLee, line 908: I suggest a new name, `claimValidatorReward` :)

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1612 @@
...
+
+                    for (uint256 tokenIdIndex = 0; tokenIdIndex < tokenId.length; ) {
+                        KGH.safeTransferFrom(address(this), msg.sender, tokenId[tokenIdIndex]);
+
+                        unchecked {
+                            ++tokenIdIndex;
+                        }
+                    }
+
+                    delete kghDelegator.pendingShares[requestTimes[i]];
```
seolaoh, line 976:
```suggestion
                    delete kghDelegator.pendingShares[requestTimes[i]];
                    delete kghDelegator.pendingKghIds[requestTimes[i]];
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1612 @@
...
+    }
+
+    /**
+     * @notice Returns the amount of KGH shares that the KGH delegator has.
+     *
+     * @param validator Address of the validator.
+     * @param delegator Address of the delegator.
+     * @param tokenId   Token Id of the KGH.
+     */
+    function getKghTotalShareBalance(
```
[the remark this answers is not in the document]
sm-stack: That will be better, thanks!

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1612 @@
...
+
+    /**
+     * @notice Initiate KRO undelegation for given validator and per-mille.
+     *
+     * @param validator Address of the validator.
+     * @param shares    The amount of shares to undelegate.
+     */
+    function initUndelegate(address validator, uint128 shares) external {
+        require(
+            shares <= _vaults[validator].kroDelegators[msg.sender].shares,
```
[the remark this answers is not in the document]
sm-stack: I agree with that, will fix it!

seolaoh (approved):
LGTM

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1625 @@
...
+        uint128 totalPendingKghShares;
+    }
+
+    /**
+     * @notice Constructs the reward information of the vault of a validator.
+     *
+     * @custom:field boostedReward                Cumulated boosted reward for KGH delegators in the vault.
+     * @custom:field validatorRewardKro           Cumulated reward for the validator.
+     * @custom:field commissionRate               Commission rate of validator.
+     * @custom:field commissionMaxChangeRate      Maximum changable commission rate at once.
```
0xHansLee, line 61:
```suggestion
     * @custom:field commissionMaxChangeRate      Maximum changeable commission rate at once.
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -0,0 +1,1625 @@
...
+
+        emit RewardClaimFinalized(msg.sender, assetsToWithdraw);
+    }
+
+    /**
+     * @notice Internal function to add base and boosted reward to the vaults of finalized output submitters.
+     *
+     * @return Whether the reward distribution is done at least once or not.
+     */
+    function _distributeReward() internal returns (bool) {
```
0xHansLee, line 1059: When the reward is distributed, the real token should be transferred to this contract by L2OO contract, right?
But for now, there is no actual vault to send the reward to this contract. If so, please add TODO mark.

0xHansLee (approved):
When the comments I left before are resolved, let's merge this PR first for `ValidatorManager` PR. And I also wanna discuss about the two things for the following PR: 
1. Why don't we unify `finalizeUndelegate` and `finalizeUndelegateKgh`?
2. `_modifyBalanceWithSlashing` is to complicated. I think it is much clearer if we separate the two cases: winner & loser. If you have any specific reason of unifying them.

PR #273, opened 2024-03-18
fix(e2e): fix challenge e2e tests not working properly

seolaoh:
# Description

Fixed below issues of challenge e2e tests.

- Due to a change in the way L1 contracts are set up in e2e, some historical variables related to guardians were being set after when the guardian should have been acting in the challenge test scenario.
- In the e2e actions test, the guardian was not checking the conditions for sending the confirmation transaction.
- In the e2e setup, the challenger was unable to get the trace because the L2 node settings were not allowed to call `kroma_getBlockTraceByNumberOrHash`.
- In e2e challenge test, the logic to check that the guardian does nothing because the deleted output is invalid was poor, so it didn't check it properly.


kroma-chain-ops/genesis/layer_one.go
```diff
@@ -140,7 +142,68 @@ func PostProcessL1DeveloperGenesis(stateDB *state.MemoryStateDB, deployments *L1
...
+
+	entry, err = layout.GetStorageLayoutEntry("_quorumNumeratorHistory")
+	if err != nil {
+		return errors.New("failed to get storage layout entry for UpgradeGovernor._quorumNumeratorHistory")
+	}
+	slot = common.BigToHash(big.NewInt(int64(entry.Slot)))
+	slot = crypto.Keccak256Hash(slot.Bytes())
+
+	beforeVal := stateDB.GetState(deployments.UpgradeGovernorProxy, slot)
+	ckptVal := make([]byte, 28)
```
Pangssu, line 165: what is the `ckpt`?

kroma-validator/guardian.go
```diff
@@ -288,7 +288,7 @@ func (g *Guardian) inspectOutput(outputIndex, fromBlock, toBlock *big.Int) {
 			retry := func() bool {
 				cCtx, cCancel := context.WithTimeout(g.ctx, g.cfg.NetworkTimeout)
 				defer cCancel()
-				isFinalized, err := g.IsOutputFinalized(cCtx, outputIndex)
+				isFinalized, err := g.isOutputFinalized(g.ctx, outputIndex)
```
Pangssu, line 291: Why didn't you use `cCtx`? If you don't need to use it, delete the definition above.

kroma-validator/guardian.go
```diff
@@ -486,8 +486,8 @@ func (g *Guardian) tryConfirmRequestDeletionTx(event *bindings.SecurityCouncilDe
 	return nil
 }
 
-func (g *Guardian) checkConfirmCondition(transactionId *big.Int, outputIndex *big.Int) (bool, error) {
-	outputFinalized, err := g.IsOutputFinalized(g.ctx, outputIndex)
+func (g *Guardian) CheckConfirmCondition(ctx context.Context, transactionId *big.Int, outputIndex *big.Int) (bool, error) {
+	outputFinalized, err := g.isOutputFinalized(ctx, outputIndex)
```
Pangssu, line 490: please add a sub context with network timeout

kroma-validator/guardian.go
```diff
@@ -496,7 +496,7 @@ func (g *Guardian) checkConfirmCondition(transactionId *big.Int, outputIndex *bi
 		return false, nil
 	}
 
-	isConfirmed, err := g.isTransactionConfirmed(transactionId)
+	isConfirmed, err := g.isTransactionConfirmed(ctx, transactionId)
```
Pangssu, line 499: please add a sub context with network timeout

op-e2e/setup.go
```diff
@@ -562,28 +562,21 @@ func (cfg SystemConfig) Start(t *testing.T, _opts ...SystemConfigOption) (*Syste
 		}
 	}
 
+	// [Kroma: START]
```
Pangssu, line 566: Why did these change?

kroma-chain-ops/genesis/layer_one.go
seolaoh: Since the historical variables in the contract are stored using `CheckpointsUpgradeable` llibrary, so I named it `ckpt`.

kroma-validator/guardian.go
seolaoh: In the function `isOutputFinalized`, there is a declaration of new context with timeout.

kroma-validator/guardian.go
seolaoh: In the function `isTransactionConfirmed`, there is a declaration of new context with timeout.

op-e2e/setup.go
seolaoh: As mentioned in description of this PR, the challenger was unable to get the trace because the L2 node settings were not allowed to call `kroma_getBlockTraceByNumberOrHash`.

kroma-validator/guardian.go
Pangssu: oh, right. I missed that.

kroma-chain-ops/genesis/layer_one.go
Pangssu: It's better to use a `checkpointVal`. It's long, but I think it's better to read.

Pangssu (approved):

kroma-chain-ops/genesis/layer_one.go
```diff
@@ -140,7 +142,68 @@ func PostProcessL1DeveloperGenesis(stateDB *state.MemoryStateDB, deployments *L1
...
+
+	layout, err = bindings.GetStorageLayout("UpgradeGovernor")
+	if err != nil {
+		return errors.New("failed to get storage layout for UpgradeGovernor")
+	}
+
+	entry, err = layout.GetStorageLayoutEntry("_quorumNumeratorHistory")
+	if err != nil {
+		return errors.New("failed to get storage layout entry for UpgradeGovernor._quorumNumeratorHistory")
+	}
```
jyc228, line 160: I suggest making this code block a function.
```go
func getStorageLayerEntity(contractName, fieldName string) (solc.StorageLayoutEntry, error) {
	layout, err := bindings.GetStorageLayout(contractName)
	if err != nil {
		return solc.StorageLayoutEntry{}, fmt.Errorf("failed to get storage layout for %s", contractName)
	}
	entry, err := layout.GetStorageLayoutEntry(fieldName)
	if err != nil {
		return solc.StorageLayoutEntry{}, fmt.Errorf("failed to get storage layout for %s.%s", contractName, fieldName)
	}
	return entry, nil
}
```

kroma-chain-ops/genesis/layer_one.go
```diff
@@ -140,7 +142,68 @@ func PostProcessL1DeveloperGenesis(stateDB *state.MemoryStateDB, deployments *L1
...
+	slot = common.BigToHash(big.NewInt(int64(entry.Slot)))
+	slot = crypto.Keccak256Hash(slot.Bytes())
+
+	beforeVal := stateDB.GetState(deployments.UpgradeGovernorProxy, slot)
+	checkpointVal := make([]byte, 28)
+	copy(checkpointVal, beforeVal[:28])
+	checkpointKey := [4]byte{}
+	checkpointKey[3] = 0x01
```
jyc228, line 168: Code duplicates lines 197-200.

kroma-chain-ops/genesis/layer_one.go
```diff
@@ -140,7 +142,68 @@ func PostProcessL1DeveloperGenesis(stateDB *state.MemoryStateDB, deployments *L1
...
+	val = common.BytesToHash(append(checkpointVal, checkpointKey[:]...))
+
+	stateDB.SetState(deployments.UpgradeGovernorProxy, slot, val)
+	afterVal := stateDB.GetState(deployments.UpgradeGovernorProxy, slot)
```
jyc228, line 172: What does this code mean?
`stateDB.SetState` alone seems like it should be enough.

kroma-chain-ops/genesis/layer_one.go
seolaoh: Yes, but I thought it could be different in future. So I remained it separately.

kroma-chain-ops/genesis/layer_one.go
seolaoh: It's because to log the difference between before value and after value below line 173.

jyc228: I understand that the values are different, but isn't that the `val` variable? I don't understand assigning the `afterVal` variable with getState.

Pangssu: I agree with Logan that using `val` is sufficient.

op-e2e/system_test.go
```diff
@@ -1496,24 +1498,49 @@ func TestChallenge(t *testing.T) {
 	}
 	cancel()
 
-	// set a timeout for security council to validate output
-	ctx, cancel = context.WithTimeout(context.Background(), 10*time.Second)
+	findTransactionId := func() *big.Int {
```
jyc228, line 1501: I can understand why you named it `findTransactionId` since it's obvious what it does in the for loop underneath, but looking at that function alone, I don't understand what transaction you're trying to find or what output you want.
Also, using the auto-generated code in bindings would eliminate the need to reference the "ValidationRequested" field as a string. 

Considering these two points, I would suggest removing findTransactionId and doing it this way.

line 1470
```go
securityCouncil, err := bindings.NewSecurityCouncil(cfg.L1Deployments.SecurityCouncilProxy, l1Client)
```

```go
if challengeStatus == chal.StatusNone {
    // check validation request tx exists and not executed
    toBlock := latestBlock(t, l1Client)
    iter, _ := securityCouncil.FilterValidationRequested(&bind.FilterOpts{End: &toBlock}, nil)
    iter.Next()
    tx, err := securityCouncil.Transactions(&bind.CallOpts{}, iter.Event.TransactionId)
    iter.Close()
    ....
}
```

op-e2e/system_test.go
```diff
@@ -1496,24 +1498,49 @@ func TestChallenge(t *testing.T) {
...
+		return new(big.Int).SetBytes(logs[0].Topics[1][:])
+	}
+
+	// set a timeout for security council to validate output (provingTimeout + buffer)
+	provingTimeout, err := colosseum.PROVINGTIMEOUT(&bind.CallOpts{})
+	ctx, cancel = context.WithTimeout(context.Background(), time.Duration(provingTimeout.Uint64()+5)*time.Second)
 	defer cancel()
 
+	numCheck := 0
 	for ; ; <-ticker.C {
```
jyc228, line 1527: even though the for loop is the last piece of code. This avoids referencing variables outside of the for loop.
```go
for numCheck := 0; ; <-ticker.C {
```

op-e2e/system_test.go
```diff
@@ -1522,15 +1549,17 @@ func TestChallenge(t *testing.T) {
 				require.Equal(t, output.Submitter, cfg.Secrets.Addresses().Challenger1)
 				outputDeleted := val.IsOutputDeleted(output.OutputRoot)
 				require.True(t, outputDeleted)
```
jyc228, line 1551: I suppose it's a matter of taste, but I think the `val.IsOutputDeleted` function is so clear that you don't need to do local variable assignments.

```go
require.True(t, val.IsOutputDeleted(output.OutputRoot))
```

sm-stack (approved):
LGTM

0xHansLee (approved):
lgtm

jyc228 (approved):

jyc228:
lgtm

PR #277, opened 2024-03-25
feat(contracts): implement `ValidatorManager` and upgrade validator system in `L2OutputOracle`

seolaoh:
# Description

Implemented `ValidatorManager` contract which manages a set of validators and selects the priority validator to submit output.

`L2OutputOracle` should upgrade validator system from using `ValidatorPool` to using `ValidatorManager`. This upgrade is not backward compatible, so should be completed with caution. Here are some considerations for the overlapping period between the two validator systems.
- For the outputs submitted last 7 days immediately before the upgrade, they should be unbonded manually.
- Before the submission period of the first output for `ValidatorManager` begins, trusted validator should register and start at `ValidatorManager`. Only trusted validator can submit the first output(`m+1`).

`Colosseum` should also upgrade validator system, so following PR will implement it.


packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -0,0 +1,497 @@
...
+        vault.reward.commissionRate = commissionRate;
+        vault.reward.commissionMaxChangeRate = commissionMaxChangeRate;
+        vault.reward.commissionRateChangedAt = uint128(block.timestamp);
+
+        _delegate(msg.sender, msg.sender, assets, false);
+
+        bool canStart = assets >= MIN_START_AMOUNT;
+        if (canStart) {
+            _validatorTree.insert(msg.sender, uint120(_reflectiveWeight(vault)));
+        }
```
sm-stack, line 175: How about this?

```suggestion
        bool canStart = assets >= MIN_START_AMOUNT;

        if (canStart) {
            _delegate(msg.sender, msg.sender, assets, true);
        } else {
            _delegate(msg.sender, msg.sender, assets, false);
        }
```

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -0,0 +1,497 @@
...
+    /**
+     * @notice Updates next priority validator address. Validators with more delegation tokens have
+     *         a higher probability of being selected. The random weight selection is based on the
+     *         last finalized output root.
+     */
+    function _updatePriorityValidator() private {
+        uint120 weightSum = startedValidatorTotalWeight();
+        uint256 latestFinalizedOutputIndex = L2_ORACLE.latestFinalizedOutputIndex();
+
+        if (weightSum > 0 && latestFinalizedOutputIndex > 0) {
```
sm-stack, line 443: Is there any reason for `latestFinalizedOutputIndex > 0` check? It seems that the `outputIndex` does not go to zero even after the `ValidatorPool` is terminated..

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -1057,15 +1092,23 @@ abstract contract AssetManager is IERC721Receiver {
...
      */
     function _distributeReward() internal returns (bool) {
-        // TODO - Query the index from L2OutputOracle
-        uint256 outputIndex = _nextRewardOutputIndex;
+        uint256 outputIndex = L2_ORACLE.latestFinalizedOutputIndex() + 1;
+        uint256 latestOutputIndex = L2_ORACLE.latestOutputIndex();
+
+        if (!L2_ORACLE.VALIDATOR_POOL().isTerminated(outputIndex)) {
+            return false;
+        }
```
sm-stack, line 1100: I think this check is meaningless, since `L2OutputOracle` already checks if `ValidatorPool` is terminated before calling `afterSubmitL2Output`, isn't it?

Pangssu:
You can use custom errors to reduce code size.

See https://ethereum.org/en/developers/tutorials/downsizing-contracts-to-fight-the-contract-size-limit/#use-custom-errors-instead-of-error-messages

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -1134,37 +1176,31 @@ abstract contract AssetManager is IERC721Receiver {
...
-        UndelegateShares memory shares
+        KghDelegatorShares memory shares
     ) internal {
-        for (uint256 i = 0; i < tokenIds.length; ) {
+        for (uint256 i = 0; i < tokenIds.length; i++) {
             vault.kghDelegators[msg.sender].pendingKghIds[block.timestamp].push(tokenIds[i]);
-
-            unchecked {
-                ++i;
-            }
```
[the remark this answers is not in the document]
seolaoh: Because the places where call this function already wrap this function with unchecked block.

packages/contracts/contracts/L1/IValidatorManager.sol
```diff
@@ -0,0 +1,122 @@
...
+     *
+     * @param validator               Address of the validator.
+     * @param started                 If the validator is started or not.
+     * @param commissionRate          The commission rate the validator sets.
+     * @param commissionMaxChangeRate The max change rate of the commission the validator sets.
+     * @param assets                  The number of assets the validator self-delegates.
+     */
+    event ValidatorRegistered(
+        address indexed validator,
+        bool indexed started,
```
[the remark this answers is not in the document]
seolaoh: For filtering only the started validator, because if a validator starts with registration, the validator does not emit started event. But maybe I think it would be unnecessary.

packages/contracts/contracts/L1/IValidatorManager.sol
```diff
@@ -0,0 +1,122 @@
...
+    function tryUnjail() external;
+
+    /**
+     * @notice Checks the eligibility to submit L2 checkpoint output during output submission.
+     *         Note that only the validator whose status is CAN_SUBMIT_OUTPUT can submit output.
+     *         This function can only be called by L2OutputOracle during output submission.
+     *
+     * @param validator Address of the output submitter.
+     */
+    function checkSubmissionEligibility(address validator) external view;
```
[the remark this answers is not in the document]
seolaoh: I think this function is not appropriate for public purpose because the action of validator will be different according to some conditions which this function checks. But if using this function to check the condition to submit output, the offchain validator can't know which action he could do. So I think offchain users should use `nextValidator()`, `getStatus()` functions, wdyt?

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -0,0 +1,497 @@
...
+        uint128 _commissionRateMinChangeSeconds,
+        uint128 _roundDurationSeconds,
+        uint128 _jailPeriodSeconds,
+        uint128 _jailThreshold
+    ) AssetManager(_constructorParams) {
+        TRUSTED_VALIDATOR = _trustedValidator;
+        COMMISSION_RATE_MIN_CHANGE_SECONDS = _commissionRateMinChangeSeconds;
+
+        // Note that this value MUST be (SUBMISSION_INTERVAL * L2_BLOCK_TIME) / 2.
+        ROUND_DURATION_SECONDS = _roundDurationSeconds;
```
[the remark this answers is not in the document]
seolaoh: Since the condition requires L2OutputOracle's variables, so it cannot be checked in constructor. If we move it to initialize function, the ROUND_DURATION_SECOND should be state variable, which increases gas cost during output submission. So how about check the condition in deploy script instead?

packages/contracts/contracts/L1/IValidatorManager.sol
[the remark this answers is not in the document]
sm-stack: Yep I don't think it can filter all started validator, so maybe it's unnecessary.

packages/contracts/contracts/L1/IValidatorManager.sol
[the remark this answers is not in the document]
sm-stack: I misunderstood the purpose of the function to be used outside since it was marked as `external`. I got what you mean.

packages/contracts/contracts/L1/ValidatorManager.sol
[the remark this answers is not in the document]
sm-stack: I understand, that would be better

packages/contracts/contracts/L1/ValidatorManager.sol
seolaoh: Because in `_delegate`, we do `tree.update`, but here we need to do `tree.insert`.

packages/contracts/contracts/L1/AssetManager.sol
seolaoh: It is necessary because it checks if `ValidatorPool` is terminated with the `latestFinalizedOutputIndex`, not the output index which `L2OutputOracle` checks it with.

packages/contracts/contracts/L1/ValidatorManager.sol
sm-stack: Got it, I missed that point.

packages/contracts/contracts/L1/AssetManager.sol
sm-stack: Okay, I got it. 

There's another question: how is the reward distributed, which is submitted right before the output index that the transition to `ValidatorManager` happens? Suppose the index of the last output of `ValidatorPool` is `N`, and when the output for index at `N + 7 days` is submitted, it will call `_distributeReward()` and return `false`, so that the last output submitter for `ValidatorPool` cannot get reward. Am I missing something?

packages/contracts/contracts/L1/IValidatorManager.sol
```diff
@@ -0,0 +1,255 @@
...
+    /**
+     * @notice Attempts to unjail a validator. Only the validator who wants to unjail can call.
+     */
+    function tryUnjail() external;
+
+    /**
+     * @notice Updates the weight tree of the validator.
+     *
+     * @param validator Address of the validator.
+     * @param tryRemove Flag to try remove the validator from weight tree.
```
sm-stack, line 226:
```suggestion
     * @param tryRemove Flag to try remove the validator from validator tree.
```

packages/contracts/contracts/L1/AssetManager.sol
seolaoh: Yes, the point what you said is to be done by manually before to distribute the first reward in `ValidatorManager`. In the description of this PR, I wrote that `For the outputs submitted last 7 days immediately before the upgrade, they should be unbonded manually.`. If the finalization period of the last output of `ValidatorPool` elapsed, we need to manually send `ValidatorPool.unbond()` transaction then `ValidatorManager` can distribute reward.

packages/contracts/contracts/L1/IValidatorManager.sol
```diff
@@ -0,0 +1,255 @@
...
+     * @param validator Address of the output submitter.
+     */
+    function checkSubmissionEligibility(address validator) external view;
+
+    /**
+     * @notice Determines who can submit the L2 checkpoint output for the current round.
+     *
+     * @return Address of the validator who can submit the L2 checkpoint output for the current
+     *         round.
+     */
```
sm-stack, line 244:
```suggestion
     * @notice Determines who can submit the L2 checkpoint output for the next round.
     *
     * @return Address of the validator who can submit the L2 checkpoint output for the next
     *         round.
     */
```

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -0,0 +1,687 @@
...
+    modifier onlyL2OutputOracle() {
+        require(
+            msg.sender == address(L2_ORACLE),
+            "ValidatorManager: Only L2OutputOracle can call this function"
+        );
+        _;
+    }
+
+    /**
+     * @notice Modifier to check if the caller is the Colosseum contract.
```
sm-stack, line 138:
```suggestion
     * @notice A modifier that only allows Colosseum contract to call.
```

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -0,0 +1,687 @@
...
+            "ValidatorManager: need to register with at least min register amount"
+        );
+
+        require(
+            commissionRate <= COMMISSION_RATE_DENOM,
+            "ValidatorManager: the max value of commission rate has been exceeded"
+        );
+
+        require(
+            commissionMaxChangeRate <= COMMISSION_RATE_DENOM,
```
sm-stack, line 203: I think this should be changed as follows
```suggestion
            commissionMaxChangeRate + commissionRate <= COMMISSION_RATE_DENOM,
```

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -0,0 +1,687 @@
...
+            require(
+                validator == _nextValidator,
+                "ValidatorManager: only the next selected validator can submit output"
+            );
+        }
+
+        require(
+            getStatus(validator) == ValidatorStatus.CAN_SUBMIT_OUTPUT,
+            "ValidatorManager: validator should satisfy the condition to submit output"
+        );
```
sm-stack, line 355: Shouldn't we add checks for the validator is `TRUSTED_VALIDATOR` here, for the output submission at the `ValidatorPool` termination? Maybe `TRUSTED_VALIDATOR` has no authority to submit output in this logic at that time.

packages/contracts/contracts/L1/AssetManager.sol
sm-stack: Oh I got it, thanks.

seolaoh:
> You can use custom errors to reduce code size.
> 
> See https://ethereum.org/en/developers/tutorials/downsizing-contracts-to-fight-the-contract-size-limit/#use-custom-errors-instead-of-error-messages

With the introduction of custom error, the code size of `ValidatorManager` and `AssetManager` is reduced by about 1 kb each! Also the gas usage when reverting. Thanks for the idea 👍

kroma-bindings/bindings/weth9.go
```diff
@@ -32,7 +32,7 @@ var (
...
...84529091529020548211156105b557600080fd5b73ffffffffffffffffffffffffffffffffffffffff841660009081526004602090815260408083203384529091529020805483900390555b73ffffffffffffffffffffffffffffffffffffffff808516600081815260036020908152604080832080548890039055938716808352918490208054870190558351868152935191937fddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef929081900390910190a35060019392505050565b3360009081526003602052604090205481111561068757600080fd5b33600081815260036020526040808220805485900390555183156108fc0291849190818181858888f193505050501580156106c6573d6000803e3d6000fd5b5060408051828152905133917f7fcf532c15f0a6db0bd6d0e038bea71d30d808c7d98cb3bf7268a95bf5081b65919081900360200190a250565b60025460ff1681565b60036020526000908152604090205481565b60018054604080516020600284861615610100027fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff0190941693909304601f8101849004840282018401909252818152929183018282801561044c5780601f106104215761010080835404028352916020019161044c565b60006107a03384846104cb565b9392505050565b60046020908152600092835260408084209091529082529020548156fea265627a7a72315820efc5f72afa7dad5e5e819df9a248fc43c598325f79ed9b9d65e74aa1a0c7578464736f6c63430005110032",
```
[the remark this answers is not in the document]
seolaoh: I don't know, but maybe it's not updated before?

packages/contracts/contracts/test/AssetManager.t.sol
```diff
@@ -337,18 +265,20 @@ contract AssetManagerTest is L2OutputOracle_Initializer {
         assetManager.delegateKgh(validator, 1);
     }
 
-    function test_initUndelegate_succeeds() external {
+    function test_initUndelegate_succeeds() public {
```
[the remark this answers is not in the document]
seolaoh: This function is used in other test functions.

packages/contracts/contracts/test/AssetManager.t.sol
```diff
@@ -500,14 +441,19 @@ contract AssetManagerTest is L2OutputOracle_Initializer {
         assetManager.initUndelegateKghBatch(validator, tokenIds);
     }
 
-    function test_initClaimValidatorReward_succeeds() external {
+    function test_initClaimValidatorReward_succeeds() public {
```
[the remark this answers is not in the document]
seolaoh: Same with above comment.

packages/contracts/contracts/L1/IValidatorManager.sol
seolaoh: Actually, it returns the address according to the round (priority or public), so `current` is more appropriate.

seolaoh:
@sm-stack @Pangssu Since I force-pushed twice, compare button above shows only the diff between the 2 force-pushes. For only the diff which I applied your comments, you can check it [here](https://github.com/kroma-network/kroma/compare/4bec497..9f11c55d740ebd5f12a927273cb6c9051f0abce8).

packages/contracts/contracts/L1/ValidatorManager.sol
seolaoh: If a validator sets `commissionRate` to 3 and `commissionMaxChangeRate` to 100, he can change the `commissionRate` to 0 and change it to 100 next time. Here the `commissionRate` means only the initial value, so I don't think it should be added to the condition.

packages/contracts/contracts/L1/ValidatorManager.sol
seolaoh: Before the `ValidatorPool` is terminated, `TRUSTED_VALIDATOR` can register and start it, then it can submit the first output after the termination. Actually I thought that `TRUSTED_VALIDATOR` is same as the other validators except that it is the only one who can submit the first output in `ValidatorManager`. `TRUSTED_VALIDATOR` also needs to register and self-delegate above `MIN_START_AMOUNT`, and needs to be added to the validator tree.

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -0,0 +1,637 @@
...
+     *
+     * @return Whether the reward distribution is done at least once or not.
+     */
+    function _distributeReward() internal returns (bool) {
+        uint256 outputIndex = L2_ORACLE.latestFinalizedOutputIndex() + 1;
+        uint256 latestOutputIndex = L2_ORACLE.latestOutputIndex();
+
+        if (!L2_ORACLE.VALIDATOR_POOL().isTerminated(outputIndex)) {
+            return false;
+        }
```
Pangssu, line 467: Is it required for overlapping periods?

kroma-bindings/bindings/weth9.go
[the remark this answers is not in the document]
Pangssu: Check your foundry and abigen version

kroma-chain-ops/genesis/layer_one_test.go
```diff
@@ -61,6 +61,10 @@ func TestBuildL1DeveloperGenesis(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, deployments.ValidatorPoolProxy, valPoolAddr)
 
+	valManAddr, err := oracle.VALIDATORMANAGER(callOpts)
+	require.NoError(t, err)
+	require.Equal(t, deployments.ValidatorManagerProxy, valManAddr)
+
```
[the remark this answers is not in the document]
seolaoh: I thought this test is to check if L1 genesis block can be built, and these lines check `L2OutputOracle`'s state variable is set properly. `L2OutputOracle` does not have `AssetManager` as a state variable.

packages/contracts/deploy/L1/008-AssetManager.ts
```diff
@@ -0,0 +1,80 @@
...
+} from '../../src/deploy-utils'
+
+const deployFn: DeployFunction = async (hre) => {
+  // TODO: change l1GovernanceTokenProxyAddress address
+  // const l1GovernanceTokenProxyAddress = await getDeploymentAddress(
+  //   hre,
+  //   'l1GovernanceTokenProxy'
+  // )
+  const l1GovernanceTokenProxyAddress =
+    '0x0B306BF915C4d645ff596e518fAf3F9669b97016'
```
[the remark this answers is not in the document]
seolaoh: Maybe I brought this address from the mint token contract PR, and it should be changed to use above lines to get the actual address after rebasing onto the PR.

packages/contracts/contracts/L1/ValidatorManager.sol
seolaoh: Yes, please see [here](https://github.com/kroma-network/kroma/pull/277#discussion_r1537520598).

kroma-bindings/bindings/weth9.go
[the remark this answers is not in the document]
seolaoh: What's your version? I downgraded abigen version from 1.12.0 to 1.10.25 as `kroma-bindings/README.md` indicates, but all the bindings are changed

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -0,0 +1,637 @@
...
+            return ValidatorStatus.NONE;
+        }
+
+        if (_jail[validator] != 0) {
+            return ValidatorStatus.IN_JAIL;
+        }
+
+        if (ASSET_MANAGER.totalValidatorKro(validator) < MIN_REGISTER_AMOUNT) {
+            return ValidatorStatus.INACTIVE;
+        }
```
0xHansLee, line 413: Is this the case that the validator register successfully, then undelegate partially?
If so, can we unite the `ACTIVE` and `INACTIVE` to `PENDING`?
When the validator is `PENDING` status, validator should `start`. For starting validator, there are conditions:
- self-delegation is greater than `MIN_REGISTER_AMOUNT`
- delegation is greater than `MIN_START_AMOUNT`

In addition with this notation, `PENDING`, we can change the work `start` to `activate`. I thinkt `activate` is better than start. wdyt? In ethereum and other validator system, also activate is used.

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -0,0 +1,637 @@
...
+    }
+
+    /**
+     * @notice Slash KRO at the vault of the validator.
+     *
+     * @param loser       Address of the loser at the challenge.
+     * @param winner      Address of the winner at the challenge.
+     * @param outputIndex The index of output challenged.
+     */
+    function slash(address loser, address winner, uint256 outputIndex) external onlyColosseum {
```
0xHansLee, line 290: We don't need to jail the loser?

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -0,0 +1,637 @@
...
+     *         last finalized output root.
+     */
+    function _updatePriorityValidator() private {
+        uint120 weightSum = startedValidatorTotalWeight();
+        uint256 latestFinalizedOutputIndex = L2_ORACLE.latestFinalizedOutputIndex();
+
+        if (weightSum > 0) {
+            Types.CheckpointOutput memory output = L2_ORACLE.getL2Output(
+                latestFinalizedOutputIndex
+            );
```
0xHansLee, line 586: Let's put inside
```suggestion
        if (weightSum > 0) {
            uint256 latestFinalizedOutputIndex = L2_ORACLE.latestFinalizedOutputIndex();
            Types.CheckpointOutput memory output = L2_ORACLE.getL2Output(
                latestFinalizedOutputIndex
            );
```

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -0,0 +1,637 @@
...
+        return _validatorTree.nodes[_validatorTree.root].weightSum;
+    }
+
+    /**
+     * @notice Internal function to add output submission rewards to the vaults of finalized output
+     *         submitters.
+     *
+     * @return Whether the reward distribution is done at least once or not.
+     */
+    function _distributeReward() internal returns (bool) {
```
0xHansLee, line 461: Why don't we make this private?

0xHansLee (commented):
Overall separation `AssetManager` from `ValidatorManager` is quite clear to me. I left some comments.

BTW, why don't we rename the `AssetManager` contract. Maybe `ValidatorVault` or `StakingManager`, `DelegationManager`?

packages/contracts/contracts/test/ValidatorManager.t.sol
```diff
@@ -0,0 +1,793 @@
...
+        assertEq(valMan.MIN_START_AMOUNT(), minStartAmount);
+        assertEq(valMan.COMMISSION_RATE_MIN_CHANGE_SECONDS(), commissionRateMinChangeSeconds);
+        assertEq(valMan.ROUND_DURATION_SECONDS(), roundDuration);
+        assertEq(valMan.JAIL_PERIOD_SECONDS(), jailPeriodSeconds);
+        assertEq(valMan.JAIL_THRESHOLD(), jailThreshold);
+        assertEq(valMan.MAX_OUTPUT_FINALIZATIONS(), maxOutputFinalizations);
+        assertEq(valMan.BASE_REWARD(), baseReward);
+    }
+
+    function test_constructor_smallMinStartAmount_reverts() external {
```
0xHansLee, line 168: Let's also add revert test when commission rate is greater than denominator.

packages/contracts/contracts/L1/ValidatorManager.sol
sm-stack: I got what you mean, the suggested one would be quite unnecessary.

packages/contracts/contracts/L1/IValidatorManager.sol
```diff
@@ -0,0 +1,310 @@
+// SPDX-License-Identifier: MIT
```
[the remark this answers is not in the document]
seolaoh: I think it's good idea, but following PR(Colosseum PR) quite depends on this branch, so I will do it in the following PR.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -418,77 +335,105 @@ abstract contract AssetManager is IERC721Receiver {
...
     /**
      * @notice Modifier to check if the vault is active.
      */
     modifier checkIsActive(address validator) {
-        require(
-            msg.sender == validator || _vaults[validator].asset.validatorKro >= MIN_REGISTER_AMOUNT,
-            "AssetManager: Vault is inactive"
-        );
+        if (
+            msg.sender != validator &&
```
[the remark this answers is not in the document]
seolaoh: Only `ValidatorManager` can call `delegateToRegister`, and `delegateToRegister` is only called in `registerValidator` function. If a validator `registerValidator` then undelegate some asset and left less than `MIN_REGISTER_AMOUNT`, the validator cannot `registerValidator` again so the only way he can re-activate his validator vault is to call `delegate`. If delete the line you mentioned, the only way will be impossible.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -697,12 +681,29 @@ abstract contract AssetManager is IERC721Receiver {
...
+     * @param assets    The amount of KRO to delegate.
+     *
+     * @return The amount of shares that the Vault would exchange for the amount of assets provided.
+     */
+    function delegateToRegister(
+        address validator,
+        uint128 assets
+    ) external onlyValidatorManager returns (uint128) {
+        uint128 shares = _delegate(validator, validator, assets, false);
+        return shares;
```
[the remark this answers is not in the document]
seolaoh: Is `ValidatorRegistered` event enough?

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -0,0 +1,637 @@
...
+
+        if (assets < MIN_REGISTER_AMOUNT) revert InsufficientAsset();
+
+        if (commissionRate > COMMISSION_RATE_DENOM) revert MaxCommissionRateExceeded();
+
+        if (commissionMaxChangeRate > COMMISSION_RATE_DENOM)
+            revert MaxCommissionChangeRateExceeded();
+
+        Validator storage validatorInfo = _validatorInfo[msg.sender];
+        validatorInfo.isInitiated = true;
```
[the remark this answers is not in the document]
seolaoh: There is no way to check if the validator is registered or not using `validatorInfo`. Because except for `isInitiated`, the other fields can be 0 even in the case validator has registered.

packages/contracts/contracts/L1/ValidatorManager.sol
seolaoh: I think the separation between `ACTIVE` and `INACTIVE` is necessary because there are some actions an inactivate validator cannot do. (1. change commission rate, 2. receive new delegation)
I think to change `start` to `activate` is good idea, but I have no idea about the words to replace `INACTIVE` and `ACTIVE`. Do you have any suggestions?

packages/contracts/contracts/L1/ValidatorManager.sol
seolaoh: I think it's good idea, but following PR(Colosseum PR) quite changes `slash` function, so I will do it in the following PR.

packages/contracts/deploy/L1/010-L2OutputOracle.ts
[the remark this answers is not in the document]
seolaoh: I thought the order of scripts should be determined by the dependency between contracts, but maybe the order doesn't matter so I will revert it. And to prevent conflict during rebasing into dev, I'll leave the place for the governance token deploy script.

kroma-chain-ops/genesis/config.go
```diff
@@ -387,6 +423,54 @@ func (d *DeployConfig) Check() error {
...
+	if d.AssetManagerKgh == (common.Address{}) {
+		return fmt.Errorf("%w: AssetManagerKgh cannot be address(0)", ErrInvalidDeployConfig)
+	}
+	if d.AssetManagerKghManager == (common.Address{}) {
+		return fmt.Errorf("%w: AssetManagerKghManager cannot be address(0)", ErrInvalidDeployConfig)
+	}
+	if d.AssetManagerUndelegationPeriod == 0 {
+		return fmt.Errorf("%w: AssetManagerUndelegationPeriod cannot be 0", ErrInvalidDeployConfig)
+	}
+	if d.AssetManagerSlashingRate == 0 {
```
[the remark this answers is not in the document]
seolaoh: Since the denominator is a constant, it does not exist in the deploy config. So I'll add the check in the constructor of `AssetManager`.

packages/contracts/contracts/test/ValidatorManager.t.sol
seolaoh: The commission rate is set when validator registers. The test you mean already exists, `test_registerValidator_largeCommissionRate_reverts`.

sm-stack (approved):
LGTM

Pangssu (approved):

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
0xHansLee: With this check (`msg.sender != validator`), validators cannot delegate themselves, right? So I thought that this line should be deleted to enable to self-delegate. But its not working with current structure (separate vaults for validator and staker). 
I think I missed something. Could you pointing out?

packages/contracts/contracts/L1/ValidatorManager.sol
0xHansLee: Ah, I see. How about these?
- INACTIVE -> PENDING
- ACTIVE -> READY
It's too hard to name it 😭

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
seolaoh: This modifier checks if the sender is not validator itself but validator is in improper status to be delegated. So if the sender is validator itself, this modifier can be passed.
I cannot understand what "separate vaults for validator and staker" means. The structure of validator vault is not changed.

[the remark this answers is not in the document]
0xHansLee: Oh, I was wrong. I confused with require statement. Sorry about that.

0xHansLee (approved):
LGTM!

packages/contracts/contracts/L1/ValidatorManager.sol
seolaoh: Hmm READY is quite ambiguous because the validator cannot do anything but delegation in READY status. Also CAN_START has redundant meaning with READY. Let's keep thinking about it.. I have no idea 😢

PR #307, opened 2024-04-18
feat: add name and symbol getter for the GovernanceToken

Pangssu:
Because GovernanceToken is an implementation of a predeployed proxy, we must use an initializer to set storage-based contents like `name` and `symbol`. However, since `name` and `symbol` are private fields, their values cannot be changed in GovernanceToken. Therefore, instead of creating an initializer, i added getter functions for name and symbol.

Additionally, it was found that the initializer was not being called during the deployment of the MintManager contract, and this has been corrected.

## Summary by CodeRabbit

- **New Features**
	- Introduced new functions to retrieve the token name and symbol in GovernanceToken.
	- Added new configuration parameters for mint management during deployment.

- **Bug Fixes**
	- Updated network configuration to correct the companion network settings.

- **Refactor**
	- Streamlined the `burn` function in GovernanceToken.
	- Overhauled contract artifacts, including adding and removing several entities.

- **Documentation**
	- Updated deployment configuration documentation to include new parameters.


kangsorang (approved):

packages/contracts/contracts/governance/GovernanceToken.sol
```diff
@@ -56,13 +56,28 @@ contract GovernanceToken is KromaMintableERC20, ERC20Burnable, ERC20Votes {
...
-    ) external override onlyBridge {
+    function burn(address _from, uint256 _amount) external override onlyBridge {
         _burn(_from, _amount);
     }
 
+    /**
+     * @notice Returns the name of token.
+     *
+     * @return The name of token.
+     */
```
seolaoh, line 67: Same to the below function `symbol()`.
```suggestion
    /**
     * @inheritdoc ERC20
     */
```

packages/contracts/deploy/L2/012-MintManager.ts
```diff
@@ -14,6 +14,10 @@ const deployFn: DeployFunction = async (hre) => {
       deployConfig.mintManagerDecayingFactor,
     ],
     isProxyImpl: true,
+    initArgs: [
+      deployConfig.mintManagerRecipients,
+      deployConfig.mintManagerShares,
+    ],
     postDeployAction: async (contract) => {
```
seolaoh, line 21: Please add this code block in `postDeployAction`
```ts
      await assertContractVariable(
        contract,
        'GOVERNANCE_TOKEN',
        predeploys.GovernanceToken
      )
```

seolaoh (approved):
LGTM

PR #318, opened 2024-05-08
feat(validator): improve retry loop when validator has not enough deposit

seolaoh:
# Description

When validator does not have enough deposit, it retries checking it continuously every 1 second. It results in too many request to L1 RPC, so changed not to retry in that case. If the validator is selected as a priority validator or in public round, it fails to submit a transaction due to gas estimation failure.

### DIFF
--- a/kroma-validator/l2_output_submitter.go
+++ b/kroma-validator/l2_output_submitter.go
@@ -10,16 +10,16 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


0xHansLee (approved):

kangsorang (approved):
LGTM

sm-stack (approved):

PR #322, opened 2024-05-20
feat(contracts): support validator system V2 from genesis

seolaoh:
# Description

In the future, we will remove the codes related to validator system V1 and might have some cases to support new chain starting with V2 from genesis. Previously, the transition codes were written with the assumption that the chain starts with V1. Therefore I modified it to support the cases starting with V2 from genesis.


Pangssu (approved):

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -279,26 +279,26 @@ contract L2OutputOracle is Initializable, ISemver {
     }
 
     /**
-     * @notice Updates the latest finalized output index. This function may only be called by the
+     * @notice Updates the next finalize output index. This function may only be called by the
```
sm-stack, line 282: How about `next finalized output index` or `next output index to be finalized`, instead of `next finalize output index`?

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -600,11 +602,11 @@ contract ValidatorManager is ISemver, IValidatorManager {
      */
     function _updatePriorityValidator() private {
         uint120 weightSum = activatedValidatorTotalWeight();
+        uint256 nextFinalizeOutputIndex = L2_ORACLE.nextFinalizeOutputIndex();
 
-        if (weightSum > 0) {
-            uint256 latestFinalizedOutputIndex = L2_ORACLE.latestFinalizedOutputIndex();
+        if (weightSum > 0 && nextFinalizeOutputIndex > 0) {
```
sm-stack, line 607: Question about this: Is this addition for the test codes?

seolaoh: No, it's because we need to select next priority validator based on finalized output.

0xHansLee (approved):

sm-stack (approved):
LGTM

PR #325, opened 2024-05-27
fix: remove validator from tree when sending to jail

seolaoh:
# Description

Once validators have been sent to jail, they should not be sent to jail again before unjail, so changed to remove the validator from the validator tree when sending to jail.


packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -460,9 +460,10 @@ contract ValidatorManager is ISemver, IValidatorManager {
     /**
      * @inheritdoc IValidatorManager
      */
-    function assertCanSubmitOutput(address validator) public view {
-        if (getStatus(validator) != ValidatorStatus.ACTIVE || inJail(validator))
-            revert ImproperValidatorStatus();
+    function assertCanSubmitOutput(address validator) public view returns (bool) {
+        if (getStatus(validator) != ValidatorStatus.ACTIVE) revert ImproperValidatorStatus();
+
+        return true;
```
seolaoh, line 466: To call this function at etherscan.

sm-stack: How about making this to return false if the status is not `ACTIVE` and make the function name to `CanSubmitOutput`? Then the experience at etherscan would be better, and also the code can be simpler at `assertCanSubmitOutput` function at `l2_output_submitter.go`. How do you think?

sm-stack (commented):
Maybe we should add `insert` logic in `tryUnjail`, if we remove the validator when it is jailed.

seolaoh:
> Maybe we should add `insert` logic in `tryUnjail`, if we remove the validator when it is jailed.

Yeah I also thought about it, but I thought it defeats the purpose of the function to add `insert` in the `unjail` function. How about changing the subcommand to optionally do `unjail` and `activate` at once?

seolaoh:
https://github.com/kroma-network/kroma/pull/325#discussion_r1615600061
I cannot reply on it, I think it's good idea to change the contract function to `CanSubmitOutput`. But in the case of l2_output_submitter.go`, I think it would be helpful to log the validator status as is when the validator does not satisfy the submission condition. wdyt?

packages/contracts/contracts/L1/ValidatorManager.sol
0xHansLee: And we can rename this function to `isActivated`.

seolaoh: I think it's good idea to change the contract function to `isActivated`. But in the case of `l2_output_submitter.go`, I think it would be helpful to log the validator status as is when the validator is not activated. wdyt?

0xHansLee:
> > Maybe we should add insert logic in tryUnjail, if we remove the validator when it is jailed.
>
> Yeah I also thought about it, but I thought it defeats the purpose of the function to add insert in the unjail function. How about changing the subcommand to optionally do unjail and activate at once?

I agree to separate `insert` and `unjail` because when a validator is jailed, the amount of delegation may be less than `MIN_ACTIVATE_AMOUNT` due to undelegation. Maybe 2 step is required to re-activate validator from unjailed status.

packages/contracts/contracts/L1/ValidatorManager.sol
0xHansLee: Yes, let's log the status of validator in `l2_output_submitter.go`.

0xHansLee (approved):
lgtm

sm-stack:
> Yeah I also thought about it, but I thought it defeats the purpose of the function to add `insert` in the `unjail` function. How about changing the subcommand to optionally do `unjail` and `activate` at once?

Could you elaborate it? I thought that if jail makes the validator to be removed from the tree, unjail should make it inserted in the tree.

> I agree to separate insert and unjail because when a validator is jailed, the amount of delegation may be less than MIN_ACTIVATE_AMOUNT due to undelegation. Maybe 2 step is required to re-activate validator from unjailed status.

I think that can be handled by checking the current status of validator, and only insert to the tree if the status is `ACTIVE`. I assume that may be cheaper than sending 2 transactions to unjail and activate the validator.

seolaoh:
> > Yeah I also thought about it, but I thought it defeats the purpose of the function to add `insert` in the `unjail` function. How about changing the subcommand to optionally do `unjail` and `activate` at once?
> 
> Could you elaborate it? I thought that if jail makes the validator to be removed from the tree, unjail should make it inserted in the tree.
> 
> > I agree to separate insert and unjail because when a validator is jailed, the amount of delegation may be less than MIN_ACTIVATE_AMOUNT due to undelegation. Maybe 2 step is required to re-activate validator from unjailed status.
> 
> I think that can be handled by checking the current status of validator, and only insert to the tree if the status is `ACTIVE`. I assume that may be cheaper than sending 2 transactions to unjail and activate the validator.

I thought it's weird to add `insert` in `unjail` function, but I agree with your comment. I'll add it.

sm-stack (approved):
LGTM

Pangssu (approved):

PR #330, opened 2024-05-31
fix(validator): change challenger and guardian to scan until head L1

seolaoh:
# Description

When using `syncStatus.CurrentL1` instead of `syncStatus.HeadL1`, it is impossible to scan between `CurrentL1` and `HeadL1`. If an asserter was off when a challenge was created, the asserter would not be able to react when turned on right after the event occurred. Guardian could also have the same issue, so fixed it also.


0xHansLee (approved):
Nice fix

Pangssu (approved):

sm-stack (approved):

PR #333, opened 2024-06-18
feat: implement governance token minting and distribution

seolaoh:
# Description

Implemented `GovernanceToken` and `MintManager` contracts. `GovernanceToken` will be minted and distributed on both L1 and L2 in TGE. `GovernanceToken` is `KromaMintableERC20`, which can be bridged between both chains via standard bridge. `MintManager` is the owner of `GovernanceToken` which can mint tokens, and it has `mint` and `distribute` function to automate the minting and distribution to several recipients. `MintManager` also has an owner who has the permission to call `mint` and `distribute`.

Note that `GovernanceToken` is proxy contract but `MintManager` is not. Additionally, `GovernanceTokenProxy` is deployed using `CREATE2` to fix the same token address across multi-chains.

### Process to Follow in TGE

- deploy
    1. deploy `L1GovernanceToken` proxy with deterministic address (w/ `mintManagerOwner` set to admin)
    2. deploy `L1MintManager` (w/ `mintManagerOwner` set to owner)
    4. deploy `L1GovernanceToken` implementation
    5. upgrade `L1GovernanceToken` proxy with ownership transfer to `L1MintManager` by `mintManagerOwner`
    6. deploy `GovernanceToken` proxy with deterministic address (w/ `mintManagerOwner` set to admin)
    7. deploy `MintManager` (w/ `mintManagerOwner` set to owner)
    8. deploy `GovernanceToken` implementation
    9. upgrade `GovernanceToken` proxy with ownership transfer to `MintManager` by `mintManagerOwner`
    10. `L1GovernanceToken.changeAdmin()`, `GovernanceToken.changeAdmin()` to zero address by `mintManagerOwner`
- L1
    1. `L1MintManager.mint()`
    2. `L1MintManager.renounceOwnershipOfToken()` by `L1MintManager` owner
    3. `L1MintManager.distribute()`
- L2
    1. `MintManager.mint()`
    2. `MintManager.renounceOwnershipOfToken()` by `MintManager` owner
    3. `MintManager.distribute()`

### Test Contracts on Holesky
- L1
  - [L1GovernanceToken](https://holesky.etherscan.io/token/0xF774455462162Bc2F6186ca323aD5D3A235F30d4)
  - [L1MintManager](https://holesky.etherscan.io/address/0x913fdBB6573Ecc606819F71E777492D540346E0e)
- L2
  - [GovernanceToken](https://blockscout.holesky.kroma.cc/address/0xF774455462162Bc2F6186ca323aD5D3A235F30d4)
  - [MintManager](https://blockscout.holesky.kroma.cc/address/0x89fE5F85e2349788F0EC98Cb3C6DcD29d574BFfa)


packages/contracts/contracts/governance/MintManager.sol
```diff
@@ -0,0 +1,136 @@
...
+
+        GOVERNANCE_TOKEN.mint(address(this), totalAmount);
+
+        minted = true;
+    }
+
+    /**
+     * @notice Only the owner is allowed to distribute the GovernanceToken to specified recipients.
+     */
+    function distribute() external onlyOwner {
```
0xHansLee, line 107: The `distribute()` should also be called only once. Do we need to check if it's already distributed or not like `minted`? I don't think it's necessary because once distributed, the balance is 0, but I'm asking just in case.

kroma-chain-ops/genesis/testdata/test-deploy-config-devnet-l1.json
```diff
@@ -53,5 +52,25 @@
   "colosseumSegmentsLengths": "2,2,3,3",
   "zkVerifierHashScalar": "0xb54142d3712f03ff5c15914a3dce82c0d8a8d537e165e2342048adf14e34aa",
   "zkVerifierM56Px": "0x175a2d1aaf07ee184a955e35f3deac123ef1088b68c38bace37056dcfc9d304d",
-  "zkVerifierM56Py": "0x150856c3d1750c2578b823cb3c7d74ed3e94c1cfece7e1bfdaaff36eea788fb9"
+  "zkVerifierM56Py": "0x150856c3d1750c2578b823cb3c7d74ed3e94c1cfece7e1bfdaaff36eea788fb9",
+  "l1GovernanceTokenProxy": "0x4200000000000000000000000000000000000064",
+  "mintManagerOwner": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
+  "l2MintManagerRecipients": [
```
0xHansLee, line 58: Let's add `l1MintManagerRecipients` and `l1MintManagerShares`.

.github/workflows/test.yml
```diff
@@ -60,6 +60,5 @@ jobs:
 
       - name: Run tests
         env:
-          OP_E2E_DISABLE_PARALLEL: true
-          OP_E2E_USE_CANYON: true
```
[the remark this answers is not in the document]
seolaoh: `OP_E2E_USE_CANYON` is not used anymore.

packages/contracts/contracts/governance/MintManager.sol
seolaoh: Yes, as you said, the call before minting will be execution reverted because the token balance of this contract is 0. Also in the case of call after distribution will be execution reverted.

kroma-chain-ops/genesis/testdata/test-deploy-config-devnet-l1.json
seolaoh: Since this file is only used to test build l2 genesis build, those configs are unnecessary. This file does not contain all deploy configs like `test-deploy-config-full.json`.

kangsorang (approved):
LGTM

0xHansLee (approved):

sm-stack (approved):
LGTM

op-e2e/bridge_test.go
```diff
@@ -110,129 +110,3 @@ func TestERC20BridgeDeposits(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, l2Balance, big.NewInt(100))
 }
-
-// TestBridgeGovernanceToken tests the L1StandardBridge bridge GovernanceToken functionality.
-func TestBridgeGovernanceToken(t *testing.T) {
```
0xHansLee, line 118: May I ask why you deleted this test?

packages/contracts/scripts/rename-deploy-scripts.ts
```diff
@@ -39,6 +41,8 @@ const L2_ORDERED_NAMES = [
   'L1FeeVault',
   'KromaMintableERC20Factory',
   'KromaMintableERC721Factory',
+  'GovernanceTokenProxy',
+  'MintManager',
```
0xHansLee, line 45: Why don't we add prefix of L2 here as well?

packages/contracts/src/deploy-config.ts
```diff
@@ -209,6 +209,16 @@ interface RequiredDeployConfig {
    * L2 : Governor voting period in block.
    */
   l2GovernorVotingPeriodBlocks: number
+
+  /**
+   * The salt to determine GovernanceTokenProxy address on L1 and L2.
+   */
+  governanceTokenProxySalt: string
```
0xHansLee, line 216: Shouldn't the salt be private and managed as an environment variable? It seems like there is room for someone to exploit this salt. For example, using the salt to deploy token contracts for scams on other L2s...

packages/contracts/deploy/L1/019-L1GovernanceToken.ts
```diff
@@ -14,13 +13,16 @@ const deployFn: DeployFunction = async (hre) => {
     'L1StandardBridgeProxy'
   )
 
-  const l1MintManagerAddress = await getDeploymentAddress(hre, 'L1MintManager')
+  const Artifact__GovernanceTokenProxy = await hre.companionNetworks[
```
0xHansLee, line 16: This means that `GovernanceTokenProxy` should be deployed on L2 first? Or, why don't we get the address of deterministic token contract from [here](https://github.com/kroma-network/kroma/pull/333/files#diff-120093f03b10b9455e845a77c01950510e03c25c7133d79c3d1a3aed678c421cR200)?

packages/contracts/scripts/rename-deploy-scripts.ts
seolaoh: It's to reduce difference from OP.

packages/contracts/src/deploy-config.ts
seolaoh: If someone exploits this salt and deploy the `Proxy` contract ahead of us, the constructor argument `admin` should be set to ours to deploy to the same address. So they cannot upgrade proxy contract without our admin key. So IMO, it's just ok to use this way to manage the salt, wdyt?

packages/contracts/deploy/L1/019-L1GovernanceToken.ts
seolaoh: Then I'll just use `L1GovernanceTokenProxy` address instead of the L2 one.

op-e2e/bridge_test.go
seolaoh: Since L2 governance token contract is converted not to be a predeploy contract, if we send the transaction to deploy it on L2 in e2e setup, some tests are broken. I thought it's unnecessary to modify all those tests to add this test, so I removed it. But if we change the default value of `enableGovernance` to false and turn it on only for the tests which need it, the issues are resolved. So I'll change it in this way.

packages/contracts/src/deploy-config.ts
0xHansLee: Ah I see. It looks ok.

0xHansLee (approved):
LGTM

packages/contracts/contracts/governance/MintManager.sol
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * @notice Only the owner is allowed to transfer the ownership of the GovernanceToken.
+     *
+     * @param newMintManager The new MintManager to own the GovernanceToken.
+     */
+    function transferOwnershipOfToken(address newMintManager) external onlyOwner {
+        GOVERNANCE_TOKEN.transferOwnership(newMintManager);
+    }
+}
```
Pangssu, line 136: Is this function required?

Pangssu (approved):

packages/contracts/contracts/governance/MintManager.sol
seolaoh: I added this function to prevent the worst case if the owner of governance token contract should be changed to new `MintManager` due to a bug. It was added because `GovernanceToken` was a predeployed contract, but now we can just use new proxy contract in that worst case. Although need to find new proxy address using new salt, do you think that it's better to just remove this function?

PR #348, opened 2024-07-18
feat: implement claimKghReward function

sm-stack:
# Description

Implement `claimKghReward` function. Includes `_claimBoostedReward` which has `isCompounding` flag.

### DIFF
--- a/packages/contracts/contracts/L1/AssetManager.sol
+++ b/packages/contracts/contracts/L1/AssetManager.sol
@@ -830,6 +830,81 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
     }
 
     /**
+     * @inheritdoc IAssetManager
+     */
+    function claimKghReward(address validator, uint256[] calldata tokenIds) external {
+        if (validator == address(0)) revert ZeroAddress();
+        if (tokenIds.length == 0) revert NotAllowedZeroInput();
+
+        Vault storage vault = _vaults[validator];
+        KghDelegator storage kghDelegator = vault.kghDelegators[msg.sender];
+        if (kghDelegator.kghNum < tokenIds.length) revert InvalidTokenIdsInput();
+
+        // Auto compounding is not applied for claim function.
+        uint128 claimedRewards = _claimBoostedReward(validator, msg.sender, tokenIds.length, false);
+
+        // Burn kroShares to claim the base reward.
+        uint128 kroSharesToBurn;
+        for (uint256 i = 0; i < tokenIds.length; ) {
+            uint128 kroShares = kghDelegator.shares[tokenIds[i]].kro -
+                _convertToKroShares(validator, KGH_MANAGER.totalKroInKgh(tokenIds[i]));
+            if (kroShares == 0) revert ShareNotExists();
+
+            unchecked {
+                kroSharesToBurn += kroShares;
+
+                ++i;
+            }
+        }
+
+        uint128 kghBaseReward = _convertToKroAssets(validator, kroSharesToBurn);
+
+        unchecked {
+            claimedRewards += kghBaseReward;
+            vault.asset.totalKroShares -= kroSharesToBurn;
+            vault.asset.totalKro -= claimedRewards;
+        }
+
+        ASSET_TOKEN.safeTransfer(msg.sender, claimedRewards);
+
+        emit KghRewardClaimed(validator, msg.sender, claimedRewards, kroSharesToBurn);
+    }
+
+    /**
+     * @notice Claim the boosted reward of the delegator.
+     *
+     * @param validator     Address of the validator.
+     * @param delegator     Address of the delegator.
+     * @param kghNum        The number of KGH to claim the boosted reward.
+     * @param isCompounding Flag to auto compound the boosted reward.
+     *
+     * @return The amount of the claimed boosted reward.
+     */
+    function _claimBoostedReward(
+        address validator,
+        address delegator,
+        uint256 kghNum,
+        bool isCompounding
+    ) internal returns (uint128) {
+        Vault storage vault = _vaults[validator];
+        KghDelegator storage kghDelegator = vault.kghDelegators[delegator];
+
+        uint128 rewardPerKghStored = vault.asset.rewardPerKghStored;
+        uint128 totalBoostedReward = kghDelegator.kghNum *
+            (rewardPerKghStored - kghDelegator.rewardPerKghPaid[delegator]);
+
+        kghDelegator.rewardPerKghPaid[delegator] = rewardPerKghStored;
+
+        if (isCompounding) {
+            _delegate(validator, delegator, totalBoostedReward);
+        }
+
+        vault.asset.boostedReward -= totalBoostedReward;
+
+        return boostedReward;
+    }
+
+    /**
      * @notice Add pending assets and shares when undelegating KRO.
      *
      * @param vault  Vault of the validator.
--- a/packages/contracts/contracts/L1/AssetManager.sol
+++ b/packages/contracts/contracts/L1/AssetManager.sol
@@ -838,10 +838,10 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
 
         Vault storage vault = _vaults[validator];
         KghDelegator storage kghDelegator = vault.kghDelegators[msg.sender];
-        if (kghDelegator.kghNum < tokenIds.length) revert InvalidTokenIdsInput();
+        if (kghDelegator.kghNum < uint128(tokenIds.length)) revert InvalidTokenIdsInput();
 
         // Auto compounding is not applied for claim function.
-        uint128 claimedRewards = _claimBoostedReward(validator, msg.sender, tokenIds.length, false);
+        uint128 claimedRewards = _claimBoostedReward(validator, msg.sender, uint128(tokenIds.length), false);
 
         // Burn kroShares to claim the base reward.
         uint128 kroSharesToBurn;
@@ -883,7 +883,7 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
     function _claimBoostedReward(
         address validator,
         address delegator,
-        uint256 kghNum,
+        uint128 kghNum,
         bool isCompounding
     ) internal returns (uint128) {
         Vault storage vault = _vaults[validator];
--- a/packages/contracts/contracts/L1/AssetManager.sol
+++ b/packages/contracts/contracts/L1/AssetManager.sol
@@ -202,6 +202,26 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
     /**
      * @inheritdoc IAssetManager
      */
+    function getKghReward(
+        address validator,
+        address delegator,
+        uint256[] calldata tokenIds
+    ) external view returns (uint128) {
+        Vault storage vault = _vaults[validator];
+        KghDelegator storage kghDelegator = vault.kghDelegators[delegator];
+
+        uint128 rewardPerKghStored = vault.asset.rewardPerKghStored;
+        uint128 totalBoostedReward = kghDelegator.kghNum *
+            (rewardPerKghStored - kghDelegator.rewardPerKghPaid);
+
+        (, uint128 kghBaseReward) = _calculateBaseRewardForKgh(validator, kghDelegator, tokenIds);
+
+        return totalBoostedReward + kghBaseReward;
+    }
+
+    /**
+     * @inheritdoc IAssetManager
+     */
     function getWithdrawAccount(address validator) external view returns (address) {
         return _vaults[validator].withdrawAccount;
     }
@@ -343,7 +363,9 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
         uint128 assets
     ) external isRegistered(validator) returns (uint128) {
         if (assets == 0) revert NotAllowedZeroInput();
+        ASSET_TOKEN.safeTransferFrom(msg.sender, address(this), assets);
         uint128 shares = _delegate(validator, msg.sender, assets);
+
         emit KroDelegated(validator, msg.sender, assets, shares);
         return shares;
     }
@@ -670,6 +692,37 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
     }
 
     /**
+     * @inheritdoc IAssetManager
+     */
+    function claimKghReward(address validator, uint256[] calldata tokenIds) external {
+        if (validator == address(0)) revert ZeroAddress();
+        if (tokenIds.length == 0) revert NotAllowedZeroInput();
+
+        Vault storage vault = _vaults[validator];
+        KghDelegator storage kghDelegator = vault.kghDelegators[msg.sender];
+        if (kghDelegator.kghNum != uint128(tokenIds.length)) revert InvalidTokenIdsInput();
+
+        uint128 claimedRewards = _claimBoostedReward(validator, msg.sender);
+        if (claimedRewards == 0) revert InsufficientAsset();
+
+        (uint128 kroSharesToBurn, uint128 kghBaseReward) = _calculateBaseRewardForKgh(
+            validator,
+            kghDelegator,
+            tokenIds
+        );
+
+        unchecked {
+            claimedRewards += kghBaseReward;
+            vault.asset.totalKroShares -= kroSharesToBurn;
+            vault.asset.totalKro -= kghBaseReward;
+        }
+
+        ASSET_TOKEN.safeTransfer(msg.sender, claimedRewards);
+
+        emit KghRewardClaimed(validator, msg.sender, claimedRewards, kroSharesToBurn);
+    }
+
+    /**
      * @notice Bond KRO from validator KRO during output submission or challenge creation. This
      *         function is only called by the ValidatorManager contract.
      *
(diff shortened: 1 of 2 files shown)


packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -816,6 +816,81 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
...
+        KghDelegator storage kghDelegator = vault.kghDelegators[msg.sender];
+        if (kghDelegator.kghNum < tokenIds.length) revert InvalidTokenIdsInput();
+
+        // Auto compounding is not applied for claim function.
+        uint128 claimedRewards = _claimBoostedReward(validator, msg.sender, tokenIds.length, false);
+
+        // Burn kroShares to claim the base reward.
+        uint128 kroSharesToBurn;
+        for (uint256 i = 0; i < tokenIds.length; ) {
+            uint128 kroShares = kghDelegator.shares[tokenIds[i]].kro -
```
seolaoh, line 836:
```suggestion
            uint128 kroShares = kghDelegator.kroShares[tokenIds[i]]-
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -816,6 +816,81 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
         }
     }
 
+    /**
+     * @inheritdoc IAssetManager
+     */
+    function claimKghReward(address validator, uint256[] calldata tokenIds) external {
+        if (validator == address(0)) revert ZeroAddress();
+        if (tokenIds.length == 0) revert NotAllowedZeroInput();
```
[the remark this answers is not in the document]
seolaoh: I think it's better to allow only the case to input all the tokenIds the delegator has delegated. Otherwise, we need to allow `tokenIds.length == 0` case if the delegator wants to claim only the boosted reward. I think it's better to restrict the case to simplify. Or it also can be a way to add `tokenIds` array in `KghDelegator`, not to input `tokenIds` in this function.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -816,6 +816,81 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
...
+
+        // Auto compounding is not applied for claim function.
+        uint128 claimedRewards = _claimBoostedReward(validator, msg.sender, tokenIds.length, false);
+
+        // Burn kroShares to claim the base reward.
+        uint128 kroSharesToBurn;
+        for (uint256 i = 0; i < tokenIds.length; ) {
+            uint128 kroShares = kghDelegator.shares[tokenIds[i]].kro -
+                _convertToKroShares(validator, KGH_MANAGER.totalKroInKgh(tokenIds[i]));
+            if (kroShares == 0) revert ShareNotExists();
```
seolaoh, line 838: `kroShares` can be 0 if the delegator does not have any claimable KRO in KGH.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -816,6 +816,81 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
...
+    function claimKghReward(address validator, uint256[] calldata tokenIds) external {
+        if (validator == address(0)) revert ZeroAddress();
+        if (tokenIds.length == 0) revert NotAllowedZeroInput();
+
+        Vault storage vault = _vaults[validator];
+        KghDelegator storage kghDelegator = vault.kghDelegators[msg.sender];
+        if (kghDelegator.kghNum < tokenIds.length) revert InvalidTokenIdsInput();
+
+        // Auto compounding is not applied for claim function.
+        uint128 claimedRewards = _claimBoostedReward(validator, msg.sender, tokenIds.length, false);
```
seolaoh: Please add revert if claimedRewards is 0. Base reward cannot be more than 0 if boosted reward is 0

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -816,6 +816,81 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
...
+                ++i;
+            }
+        }
+
+        uint128 kghBaseReward = _convertToKroAssets(validator, kroSharesToBurn);
+
+        unchecked {
+            claimedRewards += kghBaseReward;
+            vault.asset.totalKroShares -= kroSharesToBurn;
+            vault.asset.totalKro -= claimedRewards;
```
seolaoh, line 852:
```suggestion
            vault.asset.totalKro -= kghBaseReward;
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -816,6 +816,81 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
...
+     * @param delegator     Address of the delegator.
+     * @param kghNum        The number of KGH to claim the boosted reward.
+     * @param isCompounding Flag to auto compound the boosted reward.
+     *
+     * @return The amount of the claimed boosted reward.
+     */
+    function _claimBoostedReward(
+        address validator,
+        address delegator,
+        uint256 kghNum,
```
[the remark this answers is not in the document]
sm-stack: I missed this part to use it, the below code should be `uint128 totalBoostedReward = kghNum * ...` not `uint128 totalBoostedReward = kghDelegator.kghNum *`, so it's needed.

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
sm-stack: Why do we need to check `UNDELEGATION_DELAY_SECONDS` here? I thought since they are not withdrawing their KGH, it was not necessary.

Regarding the second comment, I think it depends on how frontend is implemented. I suppoesed that there is a list of KGH tokens shown there, so users can pick what KGHs they wanna claim rewards at or just claim all. If there's just only one button with 'Claim all rewards', then maybe you're right.

The last point will increase gas fees in delegating and undelegating KGHs, so I didn't see it's desirable.

packages/contracts/contracts/L1/AssetManager.sol
sm-stack: That's valid. Thanks for pointing it out!

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
0xHansLee: I also think it is not withdrawal, we don't need to check this. But I would be good to have view func that how much is the claimable reward including both base reward and boosted reward.

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
seolaoh: We need to claim all the accumulated boosted reward here always, because `rewardPerKghPaid` is updated for all the tokens, not for each token.
I think it should be removed and we need to use `kghDelegator.kghNum`

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -816,6 +816,81 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
...
+        KghDelegator storage kghDelegator = vault.kghDelegators[delegator];
+
+        uint128 rewardPerKghStored = vault.asset.rewardPerKghStored;
+        uint128 totalBoostedReward = kghDelegator.kghNum *
+            (rewardPerKghStored - kghDelegator.rewardPerKghPaid[delegator]);
+
+        kghDelegator.rewardPerKghPaid[delegator] = rewardPerKghStored;
+
+        if (isCompounding) {
+            _delegate(validator, delegator, totalBoostedReward);
```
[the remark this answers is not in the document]
sm-stack: That's valid, thanks for pointing it out :)

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
seolaoh: That makes sense. It's unnecessary to check `UNDELEGATION_DELAY_SECONDS` here.

Hmm then let's clear the cases.
1. claim only boosted reward (tokenIds.length == 0)
2. claim boosted reward and part of base reward (tokenIds.length < kghDelegator.kghNum)
3. claim boosted reward and all of base reward (tokenIds.length == kghDelegator.kghNum)

 Also, I think we need to check if the token Id is really delegated by msg.sender or not using kghDelegator.delegatedAt. Only checking the length of tokenIds array could make side effects.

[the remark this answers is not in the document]
seolaoh: > I also think it is not withdrawal, we don't need to check this. But I would be good to have view func that how much is the claimable reward including both base reward and boosted reward.

I have added `getKghReward` function in the interface. It would be nice to implement it in this PR.

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
sm-stack: Oh then we cannot do this for part of KGHs we have. I'll fix it, thanks!

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
sm-stack: Got it, then I think sticking to the case for claiming all rewards (base, boosted) will be better. If we cannot claim  portion of (base + boosted), making users claiming only boosted reward or all boosted reward + portion of base reward is not meaningful that much. I was confused at the above part of claiming portion of boosted reward.

The remaining parts you suggested look all good, so I'll add the check and `getKghReward` function.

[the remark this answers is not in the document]
seolaoh: OK, then let's confirm if all the tokenIds are provided as input, and they are really delegated by the `msg.sender`.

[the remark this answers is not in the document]
0xHansLee: > Also, I think we need to check if the token Id is really delegated by msg.sender using kghDelegator.delegatedAt. Only checking the length of tokenIds array could make side effects.

valid point.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -1144,6 +1218,36 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
...
+
+        uint128 rewardPerKghStored = vault.asset.rewardPerKghStored;
+        uint128 totalBoostedReward = kghDelegator.kghNum *
+            (rewardPerKghStored - kghDelegator.rewardPerKghPaid);
+
+        kghDelegator.rewardPerKghPaid = rewardPerKghStored;
+
+        if (isCompounding) {
+            _delegate(validator, delegator, totalBoostedReward);
+        }
```
0xHansLee, line 1246: It seems that this can be handled when calling `_claimBoostedReward()` since `totalBoostedReward` is returned.
If compounding is needed, caller can delegate the `totalBoostedReward` to validator. wdyt?

sm-stack: I didn't get what you mean. Could you elaborate it?

0xHansLee: Ok. I suggest to remove `isCompounding` flag and just return `totalBoostedReward` in this function.
Then, caller can choose to delegate the `totalBoostedReward` or not. 

At first, I thought the compounding delegation should be executed in this function, but it seems to be out of this function. 

For example, in `claimKghReward`, just transfer the `totalBoostedReward` to the delegator. On the other hand, in `delegateKgh` (would be handled in #349), make delegate with the `totalBoostedReward`. Then we don't need the `isCompounding` flang here.

seolaoh: I agree it. Just return totalBoostedReward and handle auto compounding out of this function.

Additionally, I suggest to add internal view functions to calculate boosted reward and base reward of KGH delegator each, and use the two functions in `getKghReward` and `claimKghReward`. wdyt?
Then `kghDelegator.rewardPerKghPaid = rewardPerKghStored;` this part should be handled in `claimKghReward` and `delegateKgh`.

sm-stack: So you mean that the compounding process will be executed separately in `delegateKgh`? Make sense, I'll remove it!

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -669,6 +702,48 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
...
+
+        uint128 kghBaseReward = _convertToKroAssets(validator, kroSharesToBurn);
+
+        unchecked {
+            claimedRewards += kghBaseReward;
+            vault.asset.totalKroShares -= kroSharesToBurn;
+            vault.asset.totalKro -= kghBaseReward;
+        }
+
+        if (claimedRewards == 0) revert InsufficientAsset();
```
seolaoh, line 740: How about move this line in L717? Because if totalBoostedReward is 0, we don't need to process anymore.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -199,6 +199,38 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
...
+        Vault storage vault = _vaults[validator];
+        KghDelegator storage kghDelegator = vault.kghDelegators[msg.sender];
+
+        uint128 rewardPerKghStored = vault.asset.rewardPerKghStored;
+        uint128 totalBoostedReward = kghDelegator.kghNum *
+            (rewardPerKghStored - kghDelegator.rewardPerKghPaid);
+
+        uint128 kroSharesToBurn;
+        for (uint256 i = 0; i < tokenIds.length; ) {
+            uint128 kroShares = kghDelegator.kroShares[tokenIds[i]] -
```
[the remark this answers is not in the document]
0xHansLee: Even if it is view func, I think it would be better to check as @seolaoh mentioned.

packages/contracts/contracts/L1/AssetManager.sol
0xHansLee: Because if `totalBoostedReward` is zero, it means that there is no reward to claim for base reward as well, right?

seolaoh: Yep that's right

packages/contracts/contracts/L1/AssetManager.sol
sm-stack: > Additionally, I suggest to add internal view functions to calculate boosted reward and base reward of KGH delegator each, and use the two functions in getKghReward and claimKghReward. wdyt?
Then kghDelegator.rewardPerKghPaid = rewardPerKghStored; this part should be handled in claimKghReward and delegateKgh.

I've also thought about this, but in terms of base reward, making an internal function is not desirable in terms of gas usage, since we have to check `delegateAt` inside the for loop in `claimKghReward`. Also for the boosted reward, `_claimBoostedReward` is also used in `delegateKgh`, which means an internal function to calculate the reward makes `kghDelegator.rewardPerKghPaid = rewardPerKghStored;` line repeated inside `delegateKgh` function, and I didn't see much differences in code readability imo. What do you think?

packages/contracts/contracts/L1/AssetManager.sol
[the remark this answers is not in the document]
sm-stack: That's valid. Then it seems it would be better to add an internal function to abstract calculation of base reward. I'll fix it.

packages/contracts/contracts/L1/AssetManager.sol
seolaoh: OK, it's not that critical thing and I just suggested it to reduce code redundancy between the `getKghReward` function and `claimKghReward` function.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -1144,6 +1217,31 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
...
+    /**
+     * @notice Internal function to claim the boosted reward of the delegator.
+     *
+     * @param validator     Address of the validator.
+     * @param delegator     Address of the delegator.
+     * @param isCompounding Flag to auto compound the boosted reward.
+     *
+     * @return The amount of the claimed boosted reward.
+     */
+    function _claimBoostedReward(
```
seolaoh, line 1229: How about move this function right below `_initUndelegateKghBatch`?

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -873,6 +926,38 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
...
+    /**
+     * @notice Internal function to calculate base rewards from the given validator address,
+     *         delegator struct, and token ids of KGHs.
+     *
+     * @param validator Address of the validator.
+     * @param delegator KghDelegator struct of the delegator.
+     * @param tokenIds  Array of token ids of the KGH to calculate base rewards.
+     *
+     * @return The amount of shares to undelegate.
+     * @return The amount of assets to undelegate.
```
seolaoh, line 938: Please modify it correctly.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -1197,6 +1280,31 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
         }
     }
 
+    /**
+     * @notice Internal function to claim the boosted reward of the delegator.
+     *
+     * @param validator     Address of the validator.
+     * @param delegator     Address of the delegator.
+     * @param isCompounding Flag to auto compound the boosted reward.
```
seolaoh, line 1288:
```suggestion
     * @param validator Address of the validator.
     * @param delegator Address of the delegator.
```

0xHansLee (approved):

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -1197,6 +1280,30 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
         }
     }
 
+    /**
+     * @notice Internal function to claim the boosted reward of the delegator.
+     *
+     * @param validator     Address of the validator.
+     * @param delegator     Address of the delegator.
```
seolaoh, line 1287: Can you remove the spaces?
```suggestion
     * @param validator Address of the validator.
     * @param delegator Address of the delegator.
```

seolaoh (approved):

PR #350, opened 2024-07-19
feat(contracts): implement 1-step undelegation of KRO

0xHansLee:
I will separate into several PRs for implementation of undelegation KRO, KGH, and KGH batch.

### DIFF
--- a/packages/contracts/contracts/L1/AssetManager.sol
+++ b/packages/contracts/contracts/L1/AssetManager.sol
@@ -22,11 +22,6 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
     using Uint128Math for uint128;
 
     /**
-     * @notice Virtual KRO amount per KGH.
-     */
-    uint128 public constant VKRO_PER_KGH = 100e18;
-
-    /**
      * @notice The numerator of the tax.
      */
     uint128 public constant TAX_NUMERATOR = 20;
@@ -37,7 +32,7 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
     uint128 public constant TAX_DENOMINATOR = 100;
 
     /**
-     * @notice Decimals offset for the KRO and KGH shares.
+     * @notice Decimals offset for the KRO shares.
      */
     uint128 public constant DECIMAL_OFFSET = 10 ** 6;
 
@@ -164,15 +159,19 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
     /**
      * @inheritdoc IAssetManager
      */
+    function getKghNum(address validator, address delegator) external view returns (uint128) {
+        return _vaults[validator].kghDelegators[delegator].kghNum;
+    }
+
+    /**
+     * @inheritdoc IAssetManager
+     */
     function getKghTotalShareBalance(
         address validator,
         address delegator,
         uint256 tokenId
     ) external view returns (uint128, uint128) {
-        return (
-            _vaults[validator].kghDelegators[delegator].shares[tokenId].kro,
-            _vaults[validator].kghDelegators[delegator].shares[tokenId].kgh
-        );
+        return _vaults[validator].kghDelegators[delegator].kroShares[tokenId];
     }
 
     /**
@@ -187,32 +186,6 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
      */
     function previewUndelegate(address validator, uint128 shares) external view returns (uint128) {
         return _convertToKroAssets(validator, shares);
-    }
-
-    /**
-     * @inheritdoc IAssetManager
-     */
-    function previewKghDelegate(address validator) external view returns (uint128) {
-        return _convertToKghShares(validator);
-    }
-
-    /**
-     * @inheritdoc IAssetManager
-     */
-    function previewKghUndelegate(
-        address validator,
-        address delegator,
-        uint256 tokenId
-    ) external view returns (uint128) {
-        uint128 kroInKgh = KGH_MANAGER.totalKroInKgh(tokenId);
-
-        uint128 kghAssets = _convertToKghAssets(validator, delegator, tokenId) - VKRO_PER_KGH;
-        uint128 kroAssets = _convertToKroAssets(
-            validator,
-            _vaults[validator].kghDelegators[delegator].shares[tokenId].kro
-        ) - kroInKgh;
-
-        return kghAssets + kroAssets;
     }
 
     /**
@@ -266,26 +239,13 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
      *
      * @return The reflective weight of given validator.
      */
+    // TODO: modify total weight
     function reflectiveWeight(address validator) external view returns (uint128) {
         return
             _vaults[validator].asset.totalKro +
             _vaults[validator].asset.validatorKro +
             _vaults[validator].asset.boostedReward +
             _vaults[validator].asset.validatorRewardKro;
-    }
-
-    /**
-     * @notice Returns the total amount of KGH assets held by the vault.
-     *
-     * @param validator Address of the validator.
-     *
-     * @return The total amount of KGH assets held by the vault.
-     */
-    function _totalKghAssets(address validator) internal view returns (uint128) {
-        return
-            _vaults[validator].asset.totalKgh *
-            VKRO_PER_KGH +
-            _vaults[validator].asset.boostedReward;
     }
 
     /**
@@ -297,17 +257,6 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
      */
     function _totalKroShares(address validator) internal view returns (uint128) {
         return _vaults[validator].asset.totalKroShares;
-    }
-
-    /**
-     * @notice Returns the total amount of KGH shares held by the vault.
-     *
-     * @param validator Address of the validator.
-     *
-     * @return The total amount of shares held by the validator vault.
-     */
-    function _totalKghShares(address validator) internal view returns (uint128) {
-        return _vaults[validator].asset.totalKghShares;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -241,6 +241,18 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
         return _vaults[validator].asset.validatorRewardKro;
     }
 
+    /**
+     * @inheritdoc IAssetManager
+     */
+    function canUndelegateKroAt(
```
seolaoh, line 247: How about moving this function right below `previewUndelegate`? I think it's better to locate the functions related to delegators nearly.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -975,28 +931,26 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
      * @notice Internal function to delegate KRO to the validator.
      *
      * @param validator Address of the validator.
-     * @param owner     Address of the delegator.
+     * @param delegator Address of the delegator.
      * @param assets    The amount of KRO to delegate.
+     * @param shares    The amount of shares to delegate.
      *
      * @return The amount of shares that the Vault would exchange for the amount of assets provided.
```
seolaoh, line 938:
```suggestion
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -388,16 +402,21 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
     /**
      * @inheritdoc IAssetManager
      */
-    function initUndelegate(address validator, uint128 shares) external {
+    function undelegate(address validator, uint128 shares) external {
```
seolaoh, line 405: Why not changing the input to amount? It's better to use share only for internal calculation and internal purposes, imo.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -388,16 +402,21 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
...
         if (shares == 0) revert NotAllowedZeroInput();
         if (shares > _vaults[validator].kroDelegators[msg.sender].shares)
             revert InsufficientShare();
 
         uint128 assets = _convertToKroAssets(validator, shares);
         if (assets == 0) revert InsufficientAsset();
+        if (canUndelegateKroAt(validator, msg.sender) > uint128(block.timestamp))
```
seolaoh, line 412:
```suggestion

        if (canUndelegateKroAt(validator, msg.sender) > block.timestamp)
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -1063,27 +1017,23 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
...
         uint128 shares
     ) internal {
         Vault storage vault = _vaults[validator];
 
         unchecked {
             vault.asset.totalKroShares -= shares;
-            vault.kroDelegators[owner].shares -= shares;
-
+            vault.kroDelegators[delegator].shares -= shares;
             vault.asset.totalKro -= assets;
```
seolaoh, line 1035:
```suggestion
            vault.asset.totalKro -= assets;
            vault.kroDelegators[delegator].shares -= shares;
```

packages/contracts/contracts/L1/AssetManager.sol
0xHansLee: Yeah, I agree. What do you think @sm-stack?

sm-stack: Yep I also agree, since it will reduce the complexity to calculate share at the frontend.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -388,16 +402,21 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
...
      */
-    function initUndelegate(address validator, uint128 shares) external {
+    function undelegate(address validator, uint128 shares) external {
         if (shares == 0) revert NotAllowedZeroInput();
         if (shares > _vaults[validator].kroDelegators[msg.sender].shares)
             revert InsufficientShare();
 
         uint128 assets = _convertToKroAssets(validator, shares);
         if (assets == 0) revert InsufficientAsset();
+        if (canUndelegateKroAt(validator, msg.sender) > block.timestamp)
```
sm-stack, line 412:
```suggestion
        if (canUndelegateKroAt(validator, msg.sender) > uint128(block.timestamp))
```

seolaoh: I thought it's just ok not to wrap it with uint128, is it better to wrap it in gas usage?

sm-stack: I missed the comment above, and agree that it will be okay not to wrap! I just checked about the gas, and not wrapping is slightly cheaper.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -62,9 +62,9 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
     IValidatorManager public immutable VALIDATOR_MANAGER;
 
     /**
-     * @notice Delay for the finalization of undelegation.
+     * @notice Minimum delegation period. Can be updated via upgrade.
      */
-    uint256 public immutable UNDELEGATION_PERIOD;
+    uint256 public immutable MIN_DELEGATION_PERIOD;
```
seolaoh, line 67: The constructor needs to be changed.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -118,15 +118,15 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
         IKGHManager _kghManager,
         address _securityCouncil,
         IValidatorManager _validatorManager,
-        uint128 _undelegationPeriod,
+        uint128 _minDelegationPeriod,
```
seolaoh, line 121: The comment should be changed also

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -402,13 +402,15 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
     /**
      * @inheritdoc IAssetManager
      */
-    function undelegate(address validator, uint128 shares) external {
+    function undelegate(address validator, uint128 assets) external {
```
seolaoh, line 405: Maybe we don't need the function `getKroTotalShareBalance` which was added for external purpose, wdyt?

seolaoh: Also I think we don't need `previewDelegate`, `previewUndelegate`. Instead, how about adding `getKroAssets` to fetch total asset of a delegator?

seolaoh: The interface should be changed corresponding to this function.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -402,13 +402,15 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
...
+        if (assets == 0) revert InsufficientAsset();
```
seolaoh, line 406:
```suggestion
        if (assets == 0) revert NotAllowedZeroInput();
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -402,13 +402,15 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
...
+        uint128 shares = _convertToKroShares(validator, assets);
         if (shares == 0) revert NotAllowedZeroInput();
```
seolaoh, line 408:
```suggestion
        if (shares == 0) revert InsufficientShare();
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -402,13 +402,15 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
...
         if (shares > _vaults[validator].kroDelegators[msg.sender].shares)
             revert InsufficientShare();
 
         uint128 assets = _convertToKroAssets(validator, shares);
```
seolaoh: This line should be removed

packages/contracts/contracts/L1/AssetManager.sol
0xHansLee: Yeah, we don't need `getKroTotalShareBalance`, so we can remove it. 
AFAIK, `previewDelegate` and `previewUndelegate` are interface of ERC-4626, so added. I discussed about these before with @sm-stack, and concluded that follow the interface with implementation.

seolaoh: Yeah I know that it is the interface of ERC-4626, but we actually don't extend the interface and `share` is only used internally, I thought it doesn't need to exist. wdyt?

sm-stack: Although the interfaces changed in many parts compared to ERC-4626, I think those functions have their roles for checking the status of my current delegations and further delegations. For example, combining `getKroTotalShareBalance` and `previewUndelegate` may be used to query how much assets I delegated. That may help users to check their funds are still there even if the frontend is unavailable in some reasons. How do you think about this?

seolaoh: I think it's better to just add `getKroAssets` function which do the things internally what you said (query how much assets I delegated).

0xHansLee: Why don't we have all those things? `previewUndelegate`, `previewDelegate`, and `getKroAssets`.

seolaoh: I'm still not sure if we need `previewDelegate` and `previewUndelegate`, but I think we should keep `getKroTotalShareBalance` also if we keep them.

0xHansLee: Ok. Let's just leave as is for future compatibility. I'll add `getKroAssets` in this PR.

0xHansLee: Wait, I think we cannot make `getKroAssets` to get all KRO delegated, since there is no way to get all address of validator at all in the contract. Am I missing something?

seolaoh: Right, I thought `getKroAssets(validator, delegator)`

0xHansLee: Ah, I see. I confused with `totalKroAssets`, so I thought that we don't need new one. I'll add it.

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -396,16 +417,21 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
     /**
      * @inheritdoc IAssetManager
      */
-    function initUndelegate(address validator, uint128 shares) external {
-        if (shares == 0) revert NotAllowedZeroInput();
+    function undelegate(address validator, uint128 assets) external {
```
seolaoh, line 420: Please modify the interface of this function!

packages/contracts/contracts/L1/AssetManager.sol
0xHansLee: Added in [6692470](https://github.com/kroma-network/kroma/pull/350/commits/669247060137c19dc2532db248bb709eac9a676d)

packages/contracts/contracts/L1/AssetManager.sol
0xHansLee: Oops, thanks

packages/contracts/contracts/L1/interfaces/IAssetManager.sol
```diff
@@ -507,9 +507,9 @@ interface IAssetManager {
      * @notice Undelegate the KRO of given shares for the given validator.
```
seolaoh, line 507: Please modify here also!

0xHansLee: Sorry about that. I guess I just wanted to merge quickly...

seolaoh (approved):

0xHansLee:
@sm-stack please review this PR.

sm-stack (approved):
LGTM

PR #351, opened 2024-07-19
feat(contracts): unbond validator KRO

seolaoh:
# Description

Unbond validator KRO during output finalization and slashing.

Additionally, there is a point to consider related to challenge process. I wanna discuss about the way to handle it, which I implemented in 83f1908c1a3e5f01a6659dd6c28d61ca79ae43a1.

When a malicious challenger successfully proved fault due to a ZK bug, the asserter will be slashed and not have bond anymore. But after challenge dismissed by security council, the output submitter is reverted to the asserter and the asserter's asset will be tried to be unbonded when output being finalized. In that case, the asserter does not have bond since it has been slashed, so output finalization will be reverted.

I made a change to replace the output submitter to security council when dismissing challenge, and block proving fault when submitter is security council because the output has already been validated by security council.

Please share your thoughts about the solution is appropriate or not.

### DIFF
--- a/packages/contracts/contracts/L1/AssetManager.sol
+++ b/packages/contracts/contracts/L1/AssetManager.sol
@@ -173,7 +173,7 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
     }
 
     /**
- * @inheritdoc IAssetManager
+     * @inheritdoc IAssetManager
      */
     function getKghReward(
         address validator,
@@ -689,19 +689,29 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
      * @param validator Address of the validator.
      */
     function bondValidatorKro(address validator) external onlyValidatorManager {
-        Vault storage vault = _vaults[validator];
-        if (vault.asset.validatorKro - vault.asset.validatorKroBonded < BOND_AMOUNT)
-            revert InsufficientAsset();
+        Asset storage asset = _vaults[validator].asset;
+        if (asset.validatorKro - asset.validatorKroBonded < BOND_AMOUNT) revert InsufficientAsset();
 
         unchecked {
-            vault.asset.validatorKroBonded += BOND_AMOUNT;
+            asset.validatorKroBonded += BOND_AMOUNT;
         }
 
         emit ValidatorKroBonded(
             validator,
             BOND_AMOUNT,
-            vault.asset.validatorKro - vault.asset.validatorKroBonded
+            asset.validatorKro - asset.validatorKroBonded
         );
+    }
+
+    /**
+     * @notice Unbond KRO from validator KRO during output finalization or challenge slashing. This
+     *         function is only called by the ValidatorManager contract.
+     *
+     * @param validator Address of the validator.
+     */
+    function unbondValidatorKro(address validator) external onlyValidatorManager {
+        Asset storage asset = _vaults[validator].asset;
+        _unbondValidatorKro(validator, asset);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -687,15 +708,16 @@ contract Colosseum is Initializable, ISemver {
         if (_outputRoot == DELETED_OUTPUT_ROOT) revert CannotRollbackOutputToZero();
         if (L2_ORACLE.getL2Output(_outputIndex).outputRoot != DELETED_OUTPUT_ROOT)
             revert OutputNotDeleted();
-        verifiedPublicInputs[_publicInputHash] = false;
 
         // Rollback output root.
-        L2_ORACLE.replaceL2Output(_outputIndex, _outputRoot, _asserter);
+        L2_ORACLE.replaceL2Output(_outputIndex, _outputRoot, SECURITY_COUNCIL);
```
sm-stack, line 713: As I understand, this solution to replace the submitter to security council in the case of ZK bug gives
- challenge reward to asserter
- output reward to security council
- no bond back to the asserter
Is this right? Then I think it's somewhat unfair to give the output reward to security council, since it was actually the fund that should go to the asserter if there was no ZK related bug. How about making an extra function like `recoverAsserterBond()` that recovers asserter's bond back inside the below if statement?

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -689,21 +689,31 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
      * @param validator Address of the validator.
      */
     function bondValidatorKro(address validator) external onlyValidatorManager {
-        Vault storage vault = _vaults[validator];
-        if (vault.asset.validatorKro - vault.asset.validatorKroBonded < BOND_AMOUNT)
-            revert InsufficientAsset();
+        Asset storage asset = _vaults[validator].asset;
+        if (asset.validatorKro - asset.validatorKroBonded < BOND_AMOUNT) revert InsufficientAsset();
```
sm-stack, line 693: I think saving `asset.validatorKro - asset.validatorKroBonded` in memory variable like `remainingAssets` and using it at the bottom of this function like `emit ValidatorKroBonded(..., remainingAssets - BOND_AMOUNT)` can save some gas, wdyt?

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -303,15 +303,18 @@ contract ValidatorManager is ISemver, IValidatorManager {
...
      */
     function slash(uint256 outputIndex, address winner, address loser) external onlyColosseum {
-        (uint128 tax, uint128 challengeReward) = ASSET_MANAGER.modifyBalanceWithChallenge(
-            loser,
-            0,
-            true
-        );
+        (uint128 tax, uint128 challengeReward) = ASSET_MANAGER.decreaseBalanceWithChallenge(loser);
 
         emit Slashed(outputIndex, loser, tax + challengeReward);
```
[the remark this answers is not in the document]
seolaoh: `BOND_AMOUNT` should be fetched from `ASSET_MANAGER`, and the logic to calculate challenge reward and tax can be changed in the future. I thought it could be robust regardless of the logic in `ASSET_MANAGER` in this way. wdyt?

packages/contracts/contracts/L1/Colosseum.sol
seolaoh: Yes, what you listed is almost correct, but challenge reward will also be transferred to security council.

In this way, we need to refund the asserter's slashed bond + challenge reward + output reward to the asserter manually. In V1, we also need to refund the tax to the asserter manually in dismissing challenge case.

I also considered adding `recoverAsserterBond`, but the tax amount is already tranferred to security council during slashing, so we cannot recover the asserter's bond fully. But in output finalization, the full bond amount should be unbonded, so there are also some difficulties in that way.

packages/contracts/contracts/L1/ValidatorManager.sol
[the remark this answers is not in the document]
sm-stack: That's valid, then let's leave this as it is.

packages/contracts/contracts/L1/Colosseum.sol
sm-stack: > Yes, what you listed is almost correct, but challenge reward will also be transferred to security council.

Oh yeah that's right.

How about transferring tax back to the `AssetManager` from security council in the case of `dismissChallenge`? Is there any context at the time we decide not to return the tax back to `ValidatorPool` at V1?

seolaoh: There was no context at that time, since we did not actually undergo such cases.

Hmm then how about transferring the tax at the time of output finalization? Just add `BOND_AMOUNT` to `_pendingChallengeReward` and handle it after all the challenges are concluded.

seolaoh: @0xHansLee wdyt about this issue?

sm-stack: That seems better imo, how do you think @0xHansLee ?

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -689,19 +689,26 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
...
+
+    /**
+     * @notice Unbond KRO from validator KRO during output finalization or challenge slashing. This
+     *         function is only called by the ValidatorManager contract.
+     *
+     * @param validator Address of the validator.
+     */
+    function unbondValidatorKro(address validator) external onlyValidatorManager {
+        Asset storage asset = _vaults[validator].asset;
+        _unbondValidatorKro(validator, asset);
```
0xHansLee, line 711: Let's emit event outside of `_unbondValidatorKro`. I think it would be better to have the state change happen inside the internal function, and the event emission for the associated the state change happen in the original function.

seolaoh: Then I think it's better to just remove the internal function. I'll apply it

packages/contracts/contracts/L1/Colosseum.sol
seolaoh: I applied it, please take a look.

sm-stack (approved):
LGTM

packages/contracts/contracts/L1/Colosseum.sol
0xHansLee: It's reasonable to me. 
BTW, when the challenge is dismissed, should the asserter be refunded? Once the challenge is proven, the asserter get slashed. I think asserter should be refunded the bond that already slashed.

seolaoh: Oh I missed the part. I'll add `recoverAsserterBond` function.

packages/contracts/contracts/L1/interfaces/IValidatorManager.sol
0xHansLee, line 1: Please add `SlashReverted` event

0xHansLee (approved):

PR #352, opened 2024-07-19
feat(contracts): separate jail period

sm-stack:
# Description

Separated jail period. Slightly revised a jail period calculation logic, to reflect an edge case that the non-submissions penalty and slashing penalty intercept.


packages/contracts/contracts/L1/interfaces/IValidatorManager.sol
seolaoh, line 1: Can you mention the fact that jail period is applied hardly in the comment of `slash()` function?

seolaoh (approved):
LGTM

0xHansLee (approved):

PR #363, opened 2024-07-25
test(contracts): modify `Colosseum` tests related to V2

seolaoh:
# Description

Modified some tests of `Colosseum` related to V2. Also fixed the logic that stores boosted reward in `AssetManager` to avoid divide by zero error.

### DIFF
--- a/packages/contracts/contracts/test/Colosseum.t.sol
+++ b/packages/contracts/contracts/test/Colosseum.t.sol
@@ -11,6 +11,7 @@ import { ValidatorPool } from "../L1/ValidatorPool.sol";
 import { Proxy } from "../universal/Proxy.sol";
 import { ColosseumTestData } from "./testdata/ColosseumTestData.sol";
 import { Colosseum_Initializer, ValidatorSystemUpgrade_Initializer } from "./CommonTest.t.sol";
+import { MockL2OutputOracle } from "./ValidatorManager.t.sol";
 
 // Test the implementations of the Colosseum
 contract ColosseumTest is Colosseum_Initializer {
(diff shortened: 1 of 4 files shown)


packages/contracts/contracts/test/Colosseum.t.sol
```diff
@@ -1018,6 +1047,16 @@ contract Colosseum_ValidatorSystemUpgrade_Test is Colosseum_Initializer {
         return arr;
     }
 
+    function _firstSegments() private view returns (bytes32[] memory) {
```
0xHansLee, line 1050: What does it mean of firstSegments?

0xHansLee (commented):
Could you rebase onto base branch?

packages/contracts/contracts/test/Colosseum.t.sol
seolaoh: First segments of the challenge. I added it since the code block is repeated several times.

0xHansLee: Definitely, it's good to have this function. But I wonder the "first" means. How about `_getSegments` like `_getProof`?

0xHansLee (approved):

packages/contracts/contracts/test/Colosseum.t.sol
seolaoh: Then I'll rename it `_getFirstSegments`!

0xHansLee: Ah I see. That's much clearer!

sm-stack (approved):

PR #367, opened 2024-07-31
feat(contracts): auto activate validator when deposit

seolaoh:
# Description

Activate the validator automatically when depositing if the condition is met. Also check if the commission change is initiated when finalizing commission change.


packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -213,6 +221,14 @@ contract ValidatorManager is ISemver, IValidatorManager {
         _activateValidator(msg.sender);
     }
 
+    /**
+     * @inheritdoc IValidatorManager
+     */
+    function tryActivateValidator(address validator) external onlyAssetManager {
+        if (getStatus(validator) == ValidatorStatus.READY && !inJail(validator))
```
0xHansLee, line 228: Why don't we make this as a modifier? Then, we can reuse this to `activateValidator()` func.

seolaoh: Since this function does not revert even if the condition is not met, I think it's not appropriate to use modifier here.

0xHansLee (approved):

PR #369, opened 2024-08-01
fix(validator): disable validator to be turned on without V2

seolaoh:
# Description

Disable validators to be turned on without V2 as we're currently preparing V2 to go live first. We agreed that it's over-engineering.
Also replaced the input argument `address validator` of `tryUnjail` function with `msg.sender`. It's more convenient for the validators.


0xHansLee (approved):
LGTM. We may need to inform our eco team that there is a change in interface of validator manager.

PR #370, opened 2024-08-02
fix(contracts): avoid overflow error during calculating output reward of SC

seolaoh:


packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -315,6 +315,8 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
         if (canWithdrawAt(validator) > block.timestamp) {
             revert NotElapsedMinDelegationPeriod();
         }
+        if (VALIDATOR_MANAGER.jailExpiresAt(validator) > block.timestamp)
```
0xHansLee, line 318: Could you explain why we need this here?

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -620,6 +620,10 @@ contract ValidatorManager is ISemver, IValidatorManager {
      * @return The amount of reward from commission and base reward for the validator.
      */
     function _calculateReward(address validator) internal view returns (uint128, uint128, uint128) {
+        if (validator == ASSET_MANAGER.SECURITY_COUNCIL()) {
+            return (0, 0, BASE_REWARD);
```
0xHansLee, line 624: Is there a reason to return `BASE_REWARD` to `validatorReward` and not to `baseReward`?

packages/contracts/contracts/L1/AssetManager.sol
0xHansLee: Ah, I got the point so mark resolved.

seolaoh: To prevent a validator in jail from immediately withdrawing and then registering with a new address to become active.

packages/contracts/contracts/L1/ValidatorManager.sol
seolaoh: There are no differences between two ways, baseReward is meant to be distributed to delegators, so I returned it as `validatorReward`.

0xHansLee (approved):
Thanks for your efforts

PR #372, opened 2024-08-08
docs: fix spelling issues

cratiu222:
Hello
While looking through your docs I found and fixed several spelling issues.
Hope it helps. 
Br, Christina.


0xHansLee (commented):
Thank you for your efforts, but this typo is from Optimism code base. 
We are trying to minimize the diff between Optimism and Kroma, so left the typo as is. 
So I will close this PR in Kroma, but It would be better to contribute this changes to Optimism repo first.

PR #373, opened 2024-08-22
feat(contracts): apply audit results

seolaoh:
# Description

Applied audit results. There are also some changes to reduce the code size of `Colosseum`.


packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -343,29 +358,6 @@ contract Colosseum is Initializable, ISemver {
...
-
-    /**
-     * @notice Reverts if the output of given index is already finalized.
-     *
-     * @param _outputIndex Index of the L2 checkpoint output.
-     */
-    modifier outputNotFinalized(uint256 _outputIndex) {
-        if (L2_ORACLE.isFinalized(_outputIndex)) revert OutputAlreadyFinalized();
-        _;
-    }
```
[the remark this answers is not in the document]
seolaoh: Yes, if changing from modifier to internal function, it contributes to reduce the code size.

kroma-validator/challenger.go
```diff
@@ -734,7 +734,26 @@ func (c *Challenger) IsOutputFinalized(ctx context.Context, outputIndex *big.Int
...
 	cCtx, cCancel := context.WithTimeout(ctx, c.cfg.NetworkTimeout)
 	defer cCancel()
-	return c.colosseumContract.GetChallenge(optsutils.NewSimpleCallOpts(cCtx), outputIndex, challenger)
+
+	challenge, err := c.colosseumContract.Challenges(optsutils.NewSimpleCallOpts(cCtx), outputIndex, challenger)
+	if err != nil {
+		return bindings.TypesChallenge{}, fmt.Errorf("failed to fetch challenge data: %w", err)
+	}
+
+	segments, err := c.colosseumContract.GetSegments(optsutils.NewSimpleCallOpts(cCtx), outputIndex, challenger)
```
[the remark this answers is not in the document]
seolaoh: Even though `challenges` mapping is declared to public variable, the getter function auto generated by solidity compiler returns the `Challenge` struct excluding dynamic array field `segments`. So needed to add additional function to get `segments`. Also a function returning single field rather than whole struct has less code size.

packages/contracts/contracts/L1/Colosseum.sol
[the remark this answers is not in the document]
0xHansLee: LOL

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -823,7 +841,7 @@ contract Colosseum is Initializable, ISemver {
         bytes32 _l2ToL1MessagePasserCodeHash,
         bytes32 _messagePasserStorageRoot,
         bytes32 _stateRoot
-    ) private view {
+    ) internal view {
```
Pangssu, line 844: Why did you change all private functions to internal?

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -1078,18 +1085,18 @@ contract Colosseum is Initializable, ISemver {
...
+     * @notice Returns the challenge segments corresponding to the given L2 output index and challenger.
      *
      * @param _outputIndex Index of the L2 checkpoint output.
      * @param _challenger  Address of the challenger.
      *
-     * @return The challenge data.
+     * @return The challenge segments data.
      */
-    function getChallenge(
+    function getSegments(
```
Pangssu, line 1095: It makes sense to remove this function because the `challenges` mapping is public, but I don't understand why segments getter function is needed.

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -1107,31 +1114,6 @@ contract Colosseum is Initializable, ISemver {
...
 
-    /**
-     * @notice Returns the segment length required for that turn.
-     *
-     * @param _turn The challenge turn.
-     *
-     * @return The segments length.
-     */
-    function getSegmentsLength(uint8 _turn) public view returns (uint256) {
-        if (_turn < TURN_INIT) revert InvalidTurn();
```
Pangssu, line 1118: This condition prevents invalid `turn` parameter. Is there no such case?

kroma-validator/challenger.go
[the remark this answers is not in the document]
0xHansLee: Oh I see. I didn't know that.

0xHansLee (approved):

packages/contracts/contracts/L1/Colosseum.sol
seolaoh: Please refer [this comment](https://github.com/kroma-network/kroma/pull/373#discussion_r1726460992).

kroma-validator/challenger.go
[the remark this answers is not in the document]
Pangssu: Do you mean "abigen"?

packages/contracts/contracts/L1/Colosseum.sol
seolaoh: I changed all private "view" functions to internal functions. Initially the purpose was to utilize those functions in test contract(see `Colosseum.t.sol`), otherwise duplicate codes should be added to the test contract. Also I thought if we create another contract which inherits this contract in the future, it might be useful to have those view functions internally. It does not increase any code size, and just view functions will not have any side effects in my opinion.

packages/contracts/contracts/L1/Colosseum.sol
seolaoh: I checked all the logics which utilize this function, there was no such case.

packages/contracts/contracts/L1/Colosseum.sol
Pangssu: good point

Pangssu (approved):

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -191,6 +191,9 @@ contract ValidatorManager is ISemver, IValidatorManager {
         uint8 commissionRate,
         address withdrawAccount
     ) external {
+        if (msg.sender != tx.origin) revert NotAllowedCaller();
+        if (msg.sender.code.length > 0) revert NotAllowedCaller();
```
sm-stack, line 195: It's a nit, but how about unifying these 2 lines into one?
```suggestion
        if (msg.sender != tx.origin || msg.sender.code.length > 0) revert NotAllowedCaller();
```

sm-stack (approved):
LGTM

PR #375, opened 2024-08-28
feat(contracts): withdraw remaining balance of `ValidatorRewardVault` to `ProtocolVault`

seolaoh:
# Description

Added a function to withdraw all remaining balance of `ValidatorRewardVault` to `ProtocolVault`. It can be called by anyone, so should be upgraded after all the outputs in V1 are finalized.

Fixed some wrong test codes also.


kangsorang (approved):

sm-stack (approved):
LGTM

seolaoh:
Regarding this comment, this code block has some dependencies like the upgrade schedule. Thus we decided to handle this task a while after V2 upgrade. Let's do it later by adding a permission to call the function or just upgrading with this code and reverting again.

PR #381, opened 2024-09-20
feat(contracts): enable MPT during `proveWithdrawal` in `KromaPortal`

seolaoh:
# Description

Use MPT during `proveWithdrawal` in `KromaPortal` when the `_outputRootProof` does not include `nextBlockHash`. Some library codes are brought from Optimism `v1.7.2`.

I'll make the changes to remove `nextBlockHash` from output in following PR, which will be `OutputV0` (our current version will be `KromaOutputV0`). In that PR, I'll apply MPT version `proveWithdrawal` in e2e tests.


packages/contracts/foundry.toml
```diff
@@ -1,31 +1,100 @@
...
+wrap_comments=true
+ignore = ['contracts/vendor/WETH9.sol']
+
+################################################################
+#                         PROFILE: CI                          #
+################################################################
 
 [profile.ci]
-fuzz_runs = 512
+deny_warnings = true
```
0xHansLee, line 61: Do we need this? This seems not to be in Optimism

0xHansLee (approved):

packages/contracts/foundry.toml
seolaoh: actually I brought it from Optimism `v1.7.2` tag. Do you think it's better to bring the latest version?

0xbenyun (approved):

packages/contracts/contracts/L1/KromaPortal.sol
```diff
@@ -327,10 +347,9 @@ contract KromaPortal is Initializable, ResourceMetering, ISemver {
      *
      * @param _tx Withdrawal transaction to finalize.
      */
-    function finalizeWithdrawalTransaction(Types.WithdrawalTransaction memory _tx)
-        external
```
0xbenyun, line 331: Could you please explain why the data location of the _tx parameter in the external function of Solidity is set to memory instead of calldata? I understand that calldata is preferred for read-only parameters in external functions because it consumes less gas.

seolaoh: Yeah I think you're right and it's better to use `calldata` in the respect of gas usage. But Optimism uses `memory` here, so I think it's better to remain it to reduce the diff.

PR #383, opened 2024-09-27
feat: separate `KromaOutputV0` from `OutputV0`

seolaoh:
# Description

Separated `KromaOutputV0` from `OutputV0` so that the former is used before `mpt_time`, the latter is used after `mpt_time`. Also separated `OutputWithProofResponse` from `OutputResponse`, since `OutputWithProofResponse` will not be used anymore after the transition to zkVM.

There are some codes to be uncommented after geth version updated to the state migration version. I tested `proveWithdrawal` in MPT by manually setting `kromaChainConfig.Zktrie = false`.


op-e2e/actions/user_test.go
```diff
@@ -125,6 +133,9 @@ func runCrossLayerUserTest(gt *testing.T, test hardforkScheduledTest) {
 	if test.ecotoneTime != nil {
 		require.Zero(t, uint64(*test.ecotoneTime)%uint64(dp.DeployConfig.L2BlockTime), "ecotone fork must be aligned")
 	}
+	if test.kromaMPTTime != nil {
+		require.Zero(t, uint64(*test.kromaMPTTime)%uint64(dp.DeployConfig.L2BlockTime), "kroma mpt fork must be aligned")
+	}
```
seolaoh, line 138: @Pangssu I just added this, but I wonder if kromaMPTTime should be set to exactly the timestamp of specific L2 block or can be the middle timestamp between two L2 blocks.

kangsorang:
One suggestion that just occurred to me: what if the output is differentiated by defining a version field rather than using the presence/absence of a field value?

op-e2e/actions/user_test.go
Pangssu: It must be the exact timestamp of the L2 block.

kroma-chain-ops/genesis/genesis.go
```diff
@@ -62,7 +62,9 @@ func NewL2Genesis(config *DeployConfig, block *types.Block) (*core.Genesis, erro
 		ShanghaiTime:                  config.CanyonTime(block.Time()),
 		CancunTime:                    config.EcotoneTime(block.Time()),
 		EcotoneTime:                   config.EcotoneTime(block.Time()),
-		InteropTime:                   config.InteropTime(block.Time()),
+		// TODO(seolaoh): uncomment this when geth updated
+		// KromaMPTTime:                  config.KromaMPTTime(block.Time()),
```
Pangssu, line 66: I named it `KromaMptTime`

op-node/rollup/types.go
```diff
@@ -88,6 +88,10 @@ type Config struct {
 	// Active if EcotoneTime != nil && L2 block timestamp >= *EcotoneTime, inactive otherwise.
 	EcotoneTime *uint64 `json:"ecotone_time,omitempty"`
 
+	// KromaMPTTime sets the activation time of the Kroma MPT network upgrade.
+	// Active if KromaMPTTime != nil && L2 block timestamp >= *KromaMPTTime, inactive otherwise.
+	KromaMPTTime *uint64 `json:"kroma_mpt_time,omitempty"`
```
Pangssu, line 93:
```suggestion
	KromaMptTime *uint64 `json:"mpt_time,omitempty"`
```

seolaoh:
> One suggestion that just occurred to me: what if the output is differentiated by defining a version field rather than using the presence/absence of a field value?

The version field still exists, but just omitting it means that the version is zero which we only have currently.

op-e2e/actions/user_test.go
seolaoh: It's just for my curiosity, why should it be the exact timestamp?

Pangssu (approved):

kroma-chain-ops/genesis/genesis.go
seolaoh: Why not using `KromaMPTTime` following the [go naming convention](https://go.dev/wiki/CodeReviewComments#initialisms) for initialisms?

0xbenyun (approved):

op-node/rollup/types.go
seolaoh: I thought `kroma_mpt_time` is better since it's easy to recognize that it's kroma's own config. wdyt?

kangsorang (approved):

PR #384, opened 2024-10-08
feat: `Colosseum` supports proving fault using zkVM proof

seolaoh:
# Description

For on-chain zkVM proof verificaiton, we will use `SP1Verifier` contract. According to the [docs](https://docs.succinct.xyz/onchain-verification/contract-addresses.html), there is `SP1VerifierGateway` contract deployed on each network which automatically routes the proof to the correct version verifier. For the deploy config `zkProofVerifierSP1Verifier`, we can use that address.

I separated the verification logic of public inputs and ZK proofs to a new contract `ZKProofVerifier` because of the exceeded code size of `Colosseum`. I wrote the new contract for the general purpose to verify ZK proofs regardless of whether thy are fault proofs or validity proofs.


packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -559,20 +497,18 @@ contract Colosseum is Initializable, ISemver {
...
+     * @param _zkVMProof   The public input and proof using zkVM.
      */
     function proveFault(
         uint256 _outputIndex,
         uint256 _pos,
-        Types.PublicInputProof calldata _proof,
-        uint256[] calldata _zkproof,
-        uint256[] calldata _pair
+        Types.ZKEVMProof calldata _zkEVMProof,
+        Types.ZKVMProof calldata _zkVMProof
```
Pangssu, line 511: Is it impossible to separate the functions `proveFaultZkEVM` and `proveFaultZkVM`?

packages/contracts/contracts/test/Colosseum.t.sol
```diff
@@ -55,17 +50,38 @@ contract MockColosseum is Colosseum {
...
+            _sp1Verifier,
+            _zkVMProgramVKey
+        )
+    {}
+
+    function hashZKEVMPublicInput(
         Types.PublicInputProof calldata _proof
     ) external view returns (bytes32) {
-        return _hashPublicInput(_proof.srcOutputRootProof.stateRoot, _proof.publicInput);
+        return _hashZKEVMPublicInput(_proof.srcOutputRootProof.stateRoot, _proof.publicInput);
```
kangsorang, line 77: The location of the _hashZKEVMPublicInput function has changed, shouldn't this call be a problem?

seolaoh: That's the reason why I added this new `MockZKProofVerifier` which inherits `ZKProofVerifier`.

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -822,96 +698,97 @@ contract Colosseum is Initializable, ISemver {
...
+
+        // Verify ZK proof according to the given proof type.
+        bytes32 publicInputHash;
+        if (_isZKVM) {
+            publicInputHash = ZK_PROOF_VERIFIER.verifyZKVMProof(
+                _zkVMProof,
+                srcSegment,
+                dstSegment,
+                challenge.l1Head
+            );
```
kangsorang, line 743: I think the order of the arguments passed should be changed.
```suggestion
            publicInputHash = ZK_PROOF_VERIFIER.verifyZKVMProof(
                _zkVMProof,
                challenge.l1Head,
                srcSegment,
                dstSegment
            );
```

packages/contracts/contracts/libraries/Types.sol
```diff
@@ -224,4 +228,31 @@ library Types {
...
+    /**
+     * @notice Struct representing zkEVM public input and proof.
+     *
+     * @param publicInputProof Data for verifying the public input.
+     * @param proof            Halo2 proofs composed of points and scalars.
+     *                         See https://zcash.github.io/halo2/design/implementation/proofs.html.
+     * @param pair             Aggregated multi-opening proofs and public inputs.
+     *                         (Currently only 2 public inputs)
+     */
+    struct ZKEVMProof {
```
kangsorang, line 241: It would be nice if the naming was consistently in Zk format.
I think it's less readable when it's written like TypesZKEVMProof/TypesZKVMProof.
```suggestion
    struct ZkEVMProof {
```

packages/contracts/contracts/libraries/Types.sol
```diff
@@ -224,4 +228,31 @@ library Types {
...
+    }
+
+    /**
+     * @notice Struct representing zkVM public input and proof.
+     *
+     * @param publicValues The public values concatenated.
+     *                     (Currently 3 public inputs: bytes32 srcOutputRoot, bytes32 dstOutputRoot, bytes32 l1Head)
+     * @param proofBytes   The proof of the program execution the SP1 zkVM encoded as bytes.
+     */
+    struct ZKVMProof {
```
kangsorang, line 254: Ditto

packages/contracts/contracts/libraries/Types.sol
seolaoh: I agree that it's better to read 'zkVM' than 'ZKVM', but in go codes we also use capitalized version in case of abbreviation words. So I think it's better to unify the rule in solidity codes also, wdyt?

kangsorang (approved):
LGTM

0xbenyun (approved):
LGTM

packages/contracts/contracts/libraries/Types.sol
Pangssu: ChatGPT said,

"""
In Solidity, struct names are typically written in PascalCase. Therefore, it's recommended to name your struct as ZkEvmProof or ZkVmProof. These names clearly indicate that the struct represents a proof for a zero-knowledge EVM or zero-knowledge VM, enhancing code readability and maintainability.

Example:
```solidity
struct ZkEvmProof {
    // Field definitions
}

struct ZkVmProof {
    // Field definitions
}
```

If you prefer full names over abbreviations, you could name it ZeroKnowledgeEvmProof or ZeroKnowledgeVmProof. However, since zkEVM and zkVM are commonly used abbreviations, using the concise and easily understandable ZkEvmProof or ZkVmProof is recommended.
"""

Pangssu (approved):

PR #387, opened 2024-10-30
feat(contracts): withdraw remaining balance of `ValidatorRewardVault` to `ProtocolVault`

seolaoh:
# Description

Added a function to withdraw all remaining balance of `ValidatorRewardVault` to `ProtocolVault`. It can be called by anyone, so should be upgraded after all the outputs in V1 are finalized.

Fixed some wrong test codes also.


Pangssu (approved):

packages/contracts/contracts/L2/ValidatorRewardVault.sol
```diff
@@ -144,6 +144,29 @@ contract ValidatorRewardVault is FeeVault, ISemver {
...
+        require(
+            amount >= MIN_WITHDRAWAL_AMOUNT,
+            "ValidatorRewardVault: withdrawal amount must be greater than minimum withdrawal amount"
+        );
+        require(
+            amount > 0,
+            "ValidatorRewardVault: withdrawal amount must be greater than zero"
+        );
+
+        totalProcessed += amount;
```
0xbenyun, line 162: How about applying unchecked to the line to reduce gas costs, similar to the way it's done in other functions?

kangsorang (approved):

packages/contracts/contracts/L2/ValidatorRewardVault.sol
```diff
@@ -144,6 +144,28 @@ contract ValidatorRewardVault is FeeVault, ISemver {
...
+     * @notice Withdraws all remaining contract balance excluding reserved amount to ProtocolVault.
+     *         Reverts if the withdraw amount is less than the minimum withdrawal amount.
+     */
+    function withdrawToProtocolVault() external {
+        uint256 amount = address(this).balance - totalReserved;
+        require(
+            amount >= MIN_WITHDRAWAL_AMOUNT,
+            "ValidatorRewardVault: withdrawal amount must be greater than minimum withdrawal amount"
+        );
+        require(amount > 0, "ValidatorRewardVault: withdrawal amount must be greater than zero");
```
sm-stack, line 157: Seems MIN_WITHDRAWAL_AMOUNT is zero at mainnet, so it seems like dup. Can we just leave the second require statement?

sm-stack (approved):
LGTM

0xbenyun (approved):
LGTM

PR #393, opened 2024-11-18
feat(batcher): modify batcher to immediately submit batch transaction…

0xbenyun:
## Description

To create the block at the point where the sequencer reaches KromaMPTTime, the migration must be completed up to the block immediately before the KromaMPTTime block. Since migrations are performed based on the SafeBlock, the batcher needs to be modified to immediately generate and send a batcher transaction covering all blocks up to the block just before KromaMPTTime when it encounters this block.

This PR is related to [kroma-network/go-ethereum #124](https://github.com/kroma-network/go-ethereum/pull/124).


0xbenyun:
I have read the rules in .gitmessage and tried various settings for titles and commit messages, but the semantic PR check still fails.

op-batcher/batcher/channel_manager.go
```diff
@@ -260,6 +260,10 @@ func (s *channelManager) processBlocks() error {
 		if s.currentChannel.IsFull() {
 			break
 		}
+		if s.rollupCfg.KromaMPTTime != nil && block.Time() == *s.rollupCfg.KromaMPTTime-s.rollupCfg.BlockTime {
+			s.currentChannel.channelBuilder.setFullErr(errors.New("reached the block just before KromaMPTTime"))
```
kangsorang, line 264: It would be nice to declare it in the channel builder file along with the other errors.

kangsorang: Since setFullErr is a managed function of the channelBuilder, why not handle this logic in the AddBlock of the channelBuilder?

kangsorang, line 264: Since setFullErr is a managed function of the channelBuilder, why not handle this logic in the AddBlock of the channelBuilder?

sm-stack, line 264: It's nit, but how about setting timeout instead of fullErr? It does not change the result, but seems like timeout is more natural for me.

0xbenyun: That's a good suggestion. However, since they are used with the l1 block number as input, it could cause some confusion and added complexity. Also, since these are based on fullErr after all, I think the current implementation is better

sm-stack: Agreed, then let's go with fullErr.

Pangssu (approved):

sm-stack (approved):
LGTM

seolaoh (approved):

kangsorang (approved):

PR #396, opened 2024-11-27
feat: limit submit, create challenge for first output after MPT

kangsorang:
# Description

The trie used by geth changes with the MPT transition. (Zk-trie -> MPT trie)

This causes difficulties in fault proof for output submitted after the transition block. 

To be safe, restrict the first output submission permission after the transition to trusted validator and prevent create challenge for this output.


packages/contracts/contracts/test/Colosseum.t.sol
```diff
@@ -1141,7 +1141,7 @@ contract Colosseum_ValidatorSystemUpgrade_Test is Colosseum_Initializer {
     MockL2OutputOracle mockOracle;
     uint256 internal targetOutputIndex;
 
-    function setUp() public override {
+    function setUp() public virtual override {
```
sm-stack, line 1144: Why is the `virtual` added?

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -736,4 +765,13 @@ contract ValidatorManager is ISemver, IValidatorManager {
...
     }
+
+    /**
+     * @notice Allow restrictedOutputIndex.
+     *
+     * @param _outputIndex The output index to set the limit on.
+     */
+    function setRestrictedOutputIndex(uint256 _outputIndex) public onlyTrustedValidator {
+        restrictedOutputIndex = _outputIndex;
+    }
```
sm-stack, line 776: How about making `restrictedOutputIndex` immutable and putting this into constructor? If trusted validator key is compromised, the attacker can delay the protocol, by submitting invalid output and setting this index continuously until the contract is upgraded to fix this. 

Setting it as immutable value can induce a situation that one output can still only be submitted by trusted validator even if we should revert MPT migration for some reasons, but still it would be better than the above situation. Hdyt?

packages/contracts/contracts/test/Colosseum.t.sol
```diff
@@ -1267,7 +1267,7 @@ contract Colosseum_ValidatorSystemUpgrade_Test is Colosseum_Initializer {
         return arr;
     }
 
-    function _getFirstSegments() private view returns (bytes32[] memory) {
+    function _getFirstSegments() internal view returns (bytes32[] memory) {
```
sm-stack, line 1270: Why is the visibility of this function changed?

packages/contracts/contracts/test/L2OutputOracle.t.sol
```diff
@@ -423,7 +423,7 @@ contract L2OutputOracleTest is L2OutputOracle_Initializer {
 }
 
 contract L2OutputOracle_ValidatorSystemUpgrade_Test is ValidatorSystemUpgrade_Initializer {
-    function setUp() public override {
+    function setUp() public virtual override {
```
sm-stack, line 426: Same question with above!

packages/contracts/contracts/test/ValidatorManager.t.sol
```diff
@@ -1037,4 +1037,18 @@ contract ValidatorManagerTest is ValidatorSystemUpgrade_Initializer {
         assertEq(valMgr.activatedValidatorCount(), count);
         assertEq(valMgr.activatedValidatorTotalWeight(), totalWeight + 10);
     }
+
+    function test_setRestrictedOutputIndex_TrustedValidator_succeeds() public {
```
sm-stack, line 1041:
```suggestion
    function test_setRestrictedOutputIndex_trustedValidator_succeeds() public {
```

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -387,6 +403,15 @@ contract ValidatorManager is ISemver, IValidatorManager {
         if (!isActive(validator)) revert ImproperValidatorStatus();
     }
 
+    /**
+     * @inheritdoc IValidatorManager
+     */
+    function checkCreateChallengeEligibility(uint256 outputIndex) external view onlyColosseum {
```
sm-stack, line 409: Do we need `onlyColosseum` modifier here?

packages/contracts/contracts/test/Colosseum.t.sol
kangsorang: This was made unnecessary by the change to the test contract. Updated.

packages/contracts/contracts/test/Colosseum.t.sol
kangsorang: Same as the answer above

packages/contracts/contracts/test/L2OutputOracle.t.sol
kangsorang: Same as the answer above

packages/contracts/contracts/L1/ValidatorManager.sol
kangsorang: While restrictedOutputIndex supports the ability to restrict the submission of output, there is one consideration.  
In the event of a contingency, we may want to extend the MPT time into the future and also extend the restrictedOutputIndex into the future. 
This should be as reactive as possible, so I thought of a separate setter rather than a contract upgrade.

sm-stack: What I mean was that we don't have to set `restrictedOutputIndex` again in the emergency of reverting the upgrade. Just letting the trusted validator exclusively submit in that index does not impact protocol much, and we can handle it by upgrading the contract later to update that parameter. Rather, I thought giving the permission of setting an exclusive output (any time, even he can set it multiple times) to trusted validator can be quite excessive and dangerous in the instance of hack.

But everything is tradeoff, and we can discuss and decide how to handle this.

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -129,6 +129,14 @@ contract ValidatorManager is ISemver, IValidatorManager {
      */
     mapping(uint256 => uint128) internal _pendingChallengeReward;
 
+    /**
+     * @notice Output index only allowed to submit to TrustedValidator.
+                Challenging this output is also restricted.
+     */
+    uint256 public restrictedOutputIndex;
+
+    uint256[49] private __gap;
```
Pangssu, line 138: Why did you insert gaps?

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -736,4 +765,13 @@ contract ValidatorManager is ISemver, IValidatorManager {
             _validatorInfo[validator].noSubmissionCount = 0;
         }
     }
+
+    /**
+     * @notice Allow restrictedOutputIndex.
```
Pangssu, line 770:
```suggestion
     * @notice Sets the output index that only a trusted validator can submit.
               This can only be called by a Trusted Validator.
```

packages/contracts/contracts/L1/ValidatorManager.sol
Pangssu: If you're concerned about changing `restrictedOutputIndex` multiple times, restricting it to be changed only once could be an option.

sm-stack: Agreed, that can also be an option.

packages/contracts/contracts/L1/ValidatorManager.sol
kangsorang: The restrictedOutputIndex is a temporary field for MPT transitions, so I envision removing it in a future upgrade.

packages/contracts/contracts/L1/ValidatorManager.sol
kangsorang: Yes, this is a good discussion to be cautious about restrictedOutputIndex. 
Currently, restrictedOutputIndex has two functions
- allow submissions only for trusted validators for that output index
- disallow create challenge for that output index (trusted)
I think that if there is an urgent need to change restrictedOutputIndex, the second function above (unblocking create challenge) should also work with it. Thinking about this, it seems like the way to do it via upgrade might be problematic.

Translated with www.DeepL.com/Translator (free version)

packages/contracts/contracts/L1/ValidatorManager.sol
Pangssu: Then you can store that value with randomness key.
```solidity
    bytes32 internal constant RESTRICTED_OUTPUT_INDEX_KEY = <keccak256('restrictedOutputIndex`)>;
    ...
    function setRestrictedOutputIndex(uint256 _outputIndex) public onlyTrustedValidator {
        assembly {
            sstore(RESTRICTED_OUTPUT_INDEX_KEY, _outputIndex)
        }
    }
    
    function getRestrictedOutputIndex() public view returns (uint256) {
        uint256 index;
        assembly {
            index := sload(RESTRICTED_OUTPUT_INDEX_KEY)
        }
        return index;
    }
```

packages/contracts/contracts/test/L2OutputOracle.t.sol
```diff
@@ -565,3 +565,52 @@ contract L2OutputOracleUpgradeable_Test is L2OutputOracle_Initializer {
...
+        super.setUp();
+
+        vm.prank(trusted);
+        pool.deposit{ value: trusted.balance }();
+        _registerValidator(trusted, minActivateAmount);
+
+        // Submit outputs to ValidatorPool is terminated
+        for (uint256 i; i <= terminateOutputIndex; i++) {
+            _submitL2OutputV1();
+        }
```
seolaoh, line 580: Since the first output of V2 is always for trusted validator, so for accurate test, we should pass the first V2 output
```suggestion
        }

        _submitL2OutputV2(false);
```

packages/contracts/contracts/test/L2OutputOracle.t.sol
```diff
@@ -565,3 +565,52 @@ contract L2OutputOracleUpgradeable_Test is L2OutputOracle_Initializer {
...
+    function test_submitL2Output_restricted_trustedValidator_succeeds() public {
+        uint256 nextBlockNumber = oracle.nextBlockNumber();
+        uint256 nextOutputIndex = oracle.nextOutputIndex();
+        bytes32 dummyOutputRoot = keccak256(abi.encode());
+
+        vm.prank(trusted);
+        valMgr.setRestrictedOutputIndex(nextOutputIndex);
+        assertEq(valMgr.restrictedOutputIndex(), nextOutputIndex);
+
+        warpToSubmitTime();
```
seolaoh, line 596: It's better to warp to public round, since trusted validator is always selected for the priority validator

```suggestion
        // Warp to public round
        vm.warp(oracle.nextOutputMinL2Timestamp() + roundDuration + 1);
        assertEq(valMgr.nextValidator(), trusted);

```

packages/contracts/contracts/test/L2OutputOracle.t.sol
```diff
@@ -565,3 +565,52 @@ contract L2OutputOracleUpgradeable_Test is L2OutputOracle_Initializer {
...
+    function test_submitL2Output_restricted_notTrustedValidator_reverts() public {
+        uint256 nextBlockNumber = oracle.nextBlockNumber();
+        uint256 nextOutputIndex = oracle.nextOutputIndex();
+        bytes32 dummyOutputRoot = keccak256(abi.encode());
+
+        vm.prank(trusted);
+        valMgr.setRestrictedOutputIndex(nextOutputIndex);
+        assertEq(valMgr.restrictedOutputIndex(), nextOutputIndex);
+
+        warpToSubmitTime();
```
seolaoh, line 611: It's better to test in the situation where `asserter` has the eligibility to submit output without restrictedOutputIndex.
```suggestion
        _registerValidator(asserter, minActivateAmount);

        // Warp to public round
        vm.warp(oracle.nextOutputMinL2Timestamp() + roundDuration + 1);
        assertEq(valMgr.nextValidator(), trusted);

```

packages/contracts/contracts/L1/ValidatorManager.sol
kangsorang: Yes, I get it
It would be much safer to take MPT_FIRST_OUTPUT_INDEX to immutable and upgrade it.
I'll apply the above.

seolaoh:
Could you resolve the conflict and enable test to run?

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -129,6 +129,12 @@ contract ValidatorManager is ISemver, IValidatorManager {
      */
     mapping(uint256 => uint128) internal _pendingChallengeReward;
 
+    /**
+     * @notice Output index only allowed to submit to TrustedValidator.
+               Challenges for this outputIndex are also restricted.
+     */
+    uint128 public immutable MPT_FIRST_OUTPUT_INDEX;
```
seolaoh, line 136: Move this to be with other immutable variables

seolaoh, line 136: It's better to set it uint256, since the type of outputIndex is uint256. We always check the output index with this variable every output submission, so considering the gas usage, uint256 would be better.

packages/contracts/contracts/test/Colosseum.t.sol
```diff
@@ -1553,3 +1553,179 @@ contract Colosseum_ValidatorSystemUpgrade_Test is Colosseum_Initializer {
...
+        uint256 end = targetOutput.l2BlockNumber;
+        uint256 start = end - oracle.SUBMISSION_INTERVAL();
+
+        bytes32[] memory segments = _newSegments(challenger, 1, start, end - start);
+
+        return segments;
+    }
+
+    function test_createChallenge_succeeds() public {
+        bytes32[] memory segments = _getFirstSegments();
```
sm-stack, line 1698: Maybe we can declare this at `setUp` function, so that we can use it as a global variable. Hdyt?

packages/contracts/contracts/test/Colosseum.t.sol
```diff
@@ -1553,3 +1553,179 @@ contract Colosseum_ValidatorSystemUpgrade_Test is Colosseum_Initializer {
...
+
+        // upgrade validatorManager with targetOutputIndex + 1
+        address valMgrAddress = address(valMgr);
+        constructorParams._mptFirstOutputIndex = targetOutputIndex + 1;
+        MockValidatorManager mockValMgrImpl = new MockValidatorManager(constructorParams);
+        vm.prank(multisig);
+        Proxy(payable(valMgrAddress)).upgradeTo(address(mockValMgrImpl));
+
+        vm.prank(challenger);
+        colosseum.createChallenge(targetOutputIndex, bytes32(0), 0, segments);
```
sm-stack, line 1729: How about adding an assertEq statement that checks `MPT_FIRST_OUTPUT_INDEX` and `targetOutputIndex + 1` are same?

packages/contracts/contracts/test/ValidatorManager.t.sol
```diff
@@ -1038,3 +1039,96 @@ contract ValidatorManagerTest is ValidatorSystemUpgrade_Initializer {
...
+        mockOracle.mockSetNextFinalizeOutputIndex(terminateOutputIndex + 2);
+    }
+
+    function test_submitL2Output_mptFirstOutput_privateRound_trustedValidator_succeeds() public {
+        assertEq(valMgr.nextValidator(), trusted);
+        _submitL2OutputV2(false);
+    }
+
+    function test_submitL2Output_mptFirstOutput_publicRound_trustedValidator_succeeds() public {
+        // Warp to public round since trusted validator is always selected for the priority validator
```
seolaoh, line 1100: Since asserter and trusted already registered before the first V2 output was submitted, so asserter can be selected for the priority validator proportionaly
```suggestion
        // Warp to public round
```

packages/contracts/contracts/test/ValidatorManager.t.sol
```diff
@@ -1038,3 +1039,96 @@ contract ValidatorManagerTest is ValidatorSystemUpgrade_Initializer {
...
+        vm.warp(oracle.nextOutputMinL2Timestamp() + roundDuration + 1);
+        _submitL2OutputV2(false);
+    }
+
+    function test_submitL2Output_mptFirstOutput_publicRound_notTrustedValidator_reverts() public {
+        // Warp to public round since trusted validator is always selected for the priority validator
```
seolaoh, line 1106:
```suggestion
        // Warp to public round
```

packages/contracts/contracts/test/ValidatorManager.t.sol
```diff
@@ -1038,3 +1039,96 @@ contract ValidatorManagerTest is ValidatorSystemUpgrade_Initializer {
...
+        uint256 nextBlockNumber = oracle.nextBlockNumber();
+        bytes32 outputRoot = keccak256(abi.encode(nextBlockNumber));
+        vm.startPrank(asserter, asserter);
+        vm.expectRevert(IValidatorManager.NotSelectedPriorityValidator.selector);
+        oracle.submitL2Output(outputRoot, nextBlockNumber, 0, 0);
+    }
+
+    function test_submitL2Output_mptFirstOutput_afterUpgrade_notTrustedValidator_succeeds() public {
+        uint128 newMptFirstOutputIndex = 1000;
+        // Warp to public round since trusted validator is always selected for the priority validator
```
seolaoh, line 1118:
```suggestion
        // Warp to public round
```

packages/contracts/contracts/test/Colosseum.t.sol
```diff
@@ -1553,3 +1553,179 @@ contract Colosseum_ValidatorSystemUpgrade_Test is Colosseum_Initializer {
         assertEq(assetMgr.totalValidatorKro(challenger), minActivateAmount - bondAmount);
     }
 }
+
+contract Colosseum_MptTransition_Test is Colosseum_Initializer {
+    MockColosseum mockColosseum;
+    MockZKProofVerifier mockZKProofVerifier;
+    MockL2OutputOracle mockOracle;
```
seolaoh, line 1560: I think you don't need these mock contracts, isn't it?
```suggestion
```

packages/contracts/contracts/test/ValidatorManager.t.sol
```diff
@@ -1038,3 +1039,96 @@ contract ValidatorManagerTest is ValidatorSystemUpgrade_Initializer {
...
+            l2BlockTime,
+            startingBlockNumber,
+            startingTimestamp,
+            finalizationPeriodSeconds
+        );
+        vm.prank(multisig);
+        Proxy(payable(oracleAddress)).upgradeTo(address(mockOracleImpl));
+        mockOracle = MockL2OutputOracle(oracleAddress);
+
+        address valMgrAddress = address(valMgr);
```
seolaoh, line 1067:
```suggestion
        constructorParams.mptFirstOutputIndex = mptFirstOutputIndex;
        address valMgrAddress = address(valMgr);
```

packages/contracts/contracts/test/ValidatorManager.t.sol
```diff
@@ -1038,3 +1039,96 @@ contract ValidatorManagerTest is ValidatorSystemUpgrade_Initializer {
...
+
+        // Since the first output of V2 is always for trusted validator,
+        // so for accurate test, we should pass the first V2 output
+        uint128 assets = minActivateAmount;
+        _registerValidator(asserter, assets);
+        _registerValidator(trusted, assets);
+        _submitL2OutputV2(false);
+
+        vm.warp(oracle.finalizedAt(terminateOutputIndex + 1));
+        mockOracle.mockSetNextFinalizeOutputIndex(terminateOutputIndex + 2);
```
seolaoh, line 1091:
```suggestion
        // Since the first output of V2 is always for trusted validator,
        // so for accurate test, we should pass the first V2 output
        uint128 assets = minActivateAmount;
        _registerValidator(asserter, assets);
        _registerValidator(trusted, assets);
        for (uint256 i = oracle.nextOutputIndex(); i < mptFirstOutputIndex; i++) {
            _submitL2OutputV2(false);
        }
```

packages/contracts/contracts/test/Colosseum.t.sol
```diff
@@ -1553,3 +1553,179 @@ contract Colosseum_ValidatorSystemUpgrade_Test is Colosseum_Initializer {
...
+            _trustedValidator: trusted,
+            _requiredBondAmount: requiredBondAmount,
+            _maxUnbond: maxUnbond,
+            _roundDuration: roundDuration,
+            _terminateOutputIndex: terminateOutputIndex
+        });
+        vm.prank(multisig);
+        Proxy(payable(address(pool))).upgradeTo(address(poolImpl));
+
+        // Submit outputs until ValidatorPool is terminated
```
seolaoh, line 1624:
```suggestion
        constructorParams.mptFirstOutputIndex = mptFirstOutputIndex;
        address valMgrAddress = address(valMgr);
        ValidatorManager newValMgrImpl = new ValidatorManager(constructorParams);
        vm.prank(multisig);
        Proxy(payable(valMgrAddress)).upgradeTo(address(newValMgrImpl));
        valMgr = ValidatorManager(valMgrAddress);

        // Submit outputs until ValidatorPool is terminated
```

packages/contracts/contracts/test/Colosseum.t.sol
```diff
@@ -1553,3 +1553,179 @@ contract Colosseum_ValidatorSystemUpgrade_Test is Colosseum_Initializer {
...
+        vm.prank(trusted);
+        pool.deposit{ value: trusted.balance }();
+        for (uint256 i; i <= terminateOutputIndex; i++) {
+            _submitL2OutputV1();
+        }
+
+        // Only trusted validator can submit the first output with ValidatorManager
+        _registerValidator(trusted, minActivateAmount);
+
+        // Submit invalid output as asserter
```
seolaoh, line 1634:
```suggestion
        for (uint256 i = oracle.nextOutputIndex(); i < mptFirstOutputIndex; i++) {
            _submitL2OutputV2(false);
        }

        // Submit invalid output as asserter
```

packages/contracts/contracts/test/Colosseum.t.sol
```diff
@@ -1553,3 +1553,179 @@ contract Colosseum_ValidatorSystemUpgrade_Test is Colosseum_Initializer {
         assertEq(assetMgr.totalValidatorKro(challenger), minActivateAmount - bondAmount);
     }
 }
+
+contract Colosseum_MptTransition_Test is Colosseum_Initializer {
+    MockColosseum mockColosseum;
+    MockZKProofVerifier mockZKProofVerifier;
+    MockL2OutputOracle mockOracle;
+    uint256 internal targetOutputIndex;
```
seolaoh, line 1561: Just use mptFirstOutputIndex, wdyt?
```suggestion
```

packages/contracts/contracts/test/Colosseum.t.sol
```diff
@@ -1553,3 +1553,179 @@ contract Colosseum_ValidatorSystemUpgrade_Test is Colosseum_Initializer {
...
+        address valMgrAddress = address(valMgr);
+        constructorParams._mptFirstOutputIndex = targetOutputIndex;
+        MockValidatorManager mockValMgrImpl = new MockValidatorManager(constructorParams);
+        vm.prank(multisig);
+        Proxy(payable(valMgrAddress)).upgradeTo(address(mockValMgrImpl));
+
+        vm.prank(challenger);
+        vm.expectRevert(IValidatorManager.MptFirstOutputRestricted.selector);
+        colosseum.createChallenge(targetOutputIndex, bytes32(0), 0, segments);
+    }
```
seolaoh, line 1716:
```suggestion
    function test_createChallenge_mptFirstOutputIndex_reverts() public {
        bytes32[] memory segments = _getFirstSegments();

        vm.prank(challenger);
        vm.expectRevert(IValidatorManager.MptFirstOutputRestricted.selector);
        colosseum.createChallenge(mptFirstOutputIndex, bytes32(0), 0, segments);
    }
```

packages/contracts/contracts/test/Colosseum.t.sol
kangsorang: Yes, I want to unify the common functions into an initializer contract, but there are dependencies on the setup function and member variables of the contract.

packages/contracts/deploy/L1/021-ValidatorManager.ts
```diff
@@ -47,6 +47,8 @@ const deployFn: DeployFunction = async (hre) => {
         _baseReward: hre.deployConfig.validatorManagerBaseReward,
         _minRegisterAmount: hre.deployConfig.validatorManagerMinRegisterAmount,
         _minActivateAmount: hre.deployConfig.validatorManagerMinActivateAmount,
+        _mptFirstOutputIndex:
```
seolaoh, line 50: Please add assertContractVariable for this variable below

packages/contracts/src/deploy-config.ts
```diff
@@ -111,6 +111,11 @@ interface RequiredDeployConfig {
    */
   validatorManagerMinActivateAmount: string
 
+  /**
+   * The output index to set limits on output submissions and challenges.
```
seolaoh, line 115:
```suggestion
   * The first output index after MPT transition.
```

packages/contracts/deploy-config/devnetL1-template.json
```diff
@@ -44,6 +44,7 @@
   "validatorManagerTrustedValidator": "0x70997970C51812dc3A010C7d01b50e0d17dc79C8",
   "validatorManagerMinRegisterAmount": "0x32",
   "validatorManagerMinActivateAmount": "0x64",
+  "validatorManagerMptFirstOutputIndex": "0xffff",
```
seolaoh, line 47: We already set l2GenesisKromaMPTTimeOffset to 4 in devnet (see above), so you should set it accordingly.

packages/contracts/contracts/test/Colosseum.t.sol
```diff
@@ -1553,3 +1554,170 @@ contract Colosseum_ValidatorSystemUpgrade_Test is Colosseum_Initializer {
...
+            valMgr,
+            address(colosseum),
+            submissionInterval,
+            l2BlockTime,
+            startingBlockNumber,
+            startingTimestamp,
+            finalizationPeriodSeconds
+        );
+        vm.prank(multisig);
+        Proxy(payable(oracleAddress)).upgradeTo(address(mockOracleImpl));
```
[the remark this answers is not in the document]
kangsorang: That's right, I initially imported the mock thinking I could use it when I configured the setUp by referencing ColosseumTest, but it's not needed for the existing tests, I'll remove it.

packages/contracts/deploy-config/devnetL1-template.json
kangsorang: I'm not sure what to do.
The validatorPoolTerminateOutputIndex is 32 and the validatorManagerMptFirstOutputIndex needs to be larger than that for v2 to make sense. Rather than correcting all of this, I chose to leave the actual validatorPoolTerminateOutputIndex value large enough to be meaningful. Why not inject validatorPoolTerminateOutputIndex only in the tests that need it?

packages/contracts/contracts/test/Colosseum.t.sol
```diff
@@ -1553,3 +1554,170 @@ contract Colosseum_ValidatorSystemUpgrade_Test is Colosseum_Initializer {
         assertEq(assetMgr.totalValidatorKro(challenger), minActivateAmount - bondAmount);
     }
 }
+
+contract Colosseum_MptTransition_Test is Colosseum_Initializer {
+
```
sm-stack, line 1559:
```suggestion
```

sm-stack (approved):

packages/contracts/deploy-config/devnetL1-template.json
```diff
@@ -44,6 +44,7 @@
   "validatorManagerTrustedValidator": "0x70997970C51812dc3A010C7d01b50e0d17dc79C8",
   "validatorManagerMinRegisterAmount": "0x32",
   "validatorManagerMinActivateAmount": "0x64",
+  "validatorManagerMptFirstOutputIndex": "0x0",
```
seolaoh, line 47:
```suggestion
  "validatorManagerMptFirstOutputIndex": "0x1",
```

packages/contracts/src/contract-artifacts.ts
seolaoh, line 1: Please revive it

packages/contracts/deploy-config/devnetL1-template.json
Pangssu: Why do you think it has to be 1?

kangsorang: Since the MPT transition occurs in block 2 after 4 seconds, shouldn't the first output index after the MPT transition be 0?

kangsorang: You're right, I was wrong.
output index = 0 for startingBlockNumber(=0) 
MPT first output index = 1 is correct

seolaoh (approved):

Pangssu (approved):

PR #400, opened 2024-12-12
test(e2e): add system test on withdrawal after MPT transition

sm-stack:
Added a system test for `proveWithdrawal` after MPT transition.

The previous `SendDeposit` function has always been failing to get receipt from L2 even after the transition to MPT, which I think the deposit tx is yet to be included in L2. Instead of waiting for some random number of blocks, I revised the function to take `isKromaMPT` flag.

### DIFF
--- a/op-e2e/system_test.go
+++ b/op-e2e/system_test.go
@@ -2,6 +2,7 @@ package op_e2e
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"math"
 	"math/big"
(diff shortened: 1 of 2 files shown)


0xbenyun (approved):

op-e2e/system_test.go
```diff
@@ -1344,6 +1347,180 @@ func TestWithdrawals(t *testing.T) {
 	require.Equal(t, withdrawAmount, diff)
 }
 
+// [Kroma: START]
+// TestKromaMptWithdrawals checks that a deposit and then withdrawal execution succeeds after Kroma MPT upgrade.
+// It verifies the balance changes on L1 and L2 and has to include gas fees in the balance checks.
+// It does not check that the withdrawal can be executed prior to the end of the finality period.
+func TestKromaMptWithdrawals(t *testing.T) {
```
seolaoh, line 1354: This functions seems to have lots of duplicated parts with `TestWithdrawals` function, so how about modifying to be like `TestFees`?

seolaoh (approved):

kangsorang (approved):

PR #408, opened 2024-12-18
test(e2e): replace challege testdata and enable zkVM challenge e2e test

sm-stack:
Updated challenge testdata for zkVM, and enabled zkVM challenge test at actions test.

The public input data given by SP1 prover was formatted in an unexpected way (8 bytes padded in front of each value), so I updated `ZKProofVerifier` contract to verify in a correct way.


seolaoh:
Could you remove zkVM tests in l2_challenger_test? Also please handle the challenge tests in Colosseum.t.sol (L734)

op-e2e/system_test.go
```diff
@@ -2115,20 +2115,23 @@ func TestChallengerTimeoutByGuardian(t *testing.T) {
...
 	cfg.DeployConfig.L2GenesisDeltaTimeOffset = &genesisBlock
-	cfg.DeployConfig.L2GenesisEcotoneTimeOffset = &genesisBlock
+	cfg.DeployConfig.L2GenesisEcotoneTimeOffset = &ecotoneTimeOffset
 	cfg.DeployConfig.L2GenesisKromaMPTTimeOffset = &mptTimeOffset
 	cfg.DeployConfig.L1BlockTime = 3
 	cfg.DeployConfig.L2BlockTime = 2 // same config with L2OutputOracle
 	cfg.EnableChallenge = true
 	cfg.ChallengeProofType = testdata.ZkVMType
 	cfg.NonFinalizedOutputs = true // speed up the time till we see checkpoint outputs
+	cfg.ValidatorVersion = valhelper.ValidatorV2
```
seolaoh, line 2134: How about removing this part according to the deploy config changes (revert validatorPoolTerminateIndex to original)?

op-e2e/system_test.go
```diff
@@ -2140,6 +2143,10 @@ func TestZkVMChallenge(t *testing.T) {
 	// Deposit to ValidatorPool to be a challenger
 	validatorHelper.DepositToValPool(cfg.Secrets.Challenger1, big.NewInt(1_000_000_000))
 
+	// Register to ValidatorManager to be a challenger
+	validatorHelper.RegisterToValMgr(cfg.Secrets.Challenger1,
+		cfg.DeployConfig.ValidatorManagerMinActivateAmount.ToInt(), cfg.Secrets.Addresses().Challenger1)
```
seolaoh, line 2148:
```suggestion
```

op-e2e/testdata/challenge_test_data.go
```diff
@@ -91,22 +86,13 @@ var (
...
+		o.OutputRoot = eth.Bytes32(zkVMPrevOutput)
+		// Other values than output root won't be used in zkVM challenge, so we can set them to empty hash
+		o.WithdrawalStorageRoot = EmptyHash
+		o.StateRoot = EmptyHash
 		o.BlockRef = eth.L2BlockRef{
 			Number:     TargetBlockNumber - 1,
-			Hash:       zkVMPrevBlockhash,
+			Hash:       EmptyHash,
 			ParentHash: common.HexToHash("0x"),
 		}
```
seolaoh, line 97:
```suggestion
		// Other values than output root won't be used in zkVM challenge, so we don't have to mock them
```

op-e2e/testdata/challenge_test_data.go
```diff
@@ -187,23 +173,14 @@ func SetPrevOutputWithProofResponse(o *eth.OutputWithProofResponse) error {
...
+		// Other values than output root won't be used in zkVM challenge, so we can set them to empty hash
+		o.WithdrawalStorageRoot = EmptyHash
+		o.StateRoot = EmptyHash
 		o.BlockRef = eth.L2BlockRef{
 			Number:     TargetBlockNumber,
-			Hash:       zkVMTargetBlockhash,
-			ParentHash: zkVMPrevBlockhash,
+			Hash:       EmptyHash,
+			ParentHash: EmptyHash,
 		}
```
seolaoh, line 184:
```suggestion
		// Other values than output root won't be used in zkVM challenge, so we don't have to mock them
```

packages/contracts/contracts/L1/ZKProofVerifier.sol
```diff
@@ -200,11 +200,11 @@ contract ZKProofVerifier is ISemver {
         _validatePublicInputOutput(
             _storedSrcOutput,
             _storedDstOutput,
-            bytes32(_zkVmProof.publicValues[:32]),
-            bytes32(_zkVmProof.publicValues[32:64])
+            bytes32(_zkVmProof.publicValues[8:40]),
```
seolaoh, line 203: How about removing the prefix in challenger client?

seolaoh: Check the length of `publicValues`, if it is longer than 96, then check if each value has the specific prefix and remove the prefix? Also I think it's better to add a require phrase in verifyZkVmProof contract function, wdyt?

seolaoh: Or it's good to check it at the front of proveFaultWithZkVm function to save gas

sm-stack: I've tested it with the public values without the prefix, but the verification fails in the SP1Verifier side. I think they need the prefix for the successful verification.

seolaoh: Oh really? I understand it, thanks!

seolaoh: Then could you add some comments here?

0xbenyun (approved):

op-e2e/testdata/challenge_test_data.go
seolaoh: Could you apply the comment also?

op-e2e/testdata/challenge_test_data.go
seolaoh: Could you apply the comment also?

seolaoh (approved):

Pangssu (approved):

PR #418, opened 2025-01-20
feat: ignore KromaMPT block payload

Pangssu:
# Description

Prevents KromaMPT block from being handled as unsafe payload over P2P. 
This ensures that nodes can continue the migration process, even if they receive a new unsafe payload from the sequencer before receiving the batch containing the KromaMPT parent block, allowing them to still process the safe attribute.

### DIFF
--- a/op-batcher/batcher/channel_manager_test.go
+++ b/op-batcher/batcher/channel_manager_test.go
@@ -7,16 +7,17 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/stretchr/testify/require"
+
 	"github.com/ethereum-optimism/optimism/op-batcher/metrics"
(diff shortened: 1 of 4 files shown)


op-batcher/batcher/channel_manager_test.go
```diff
@@ -524,27 +525,17 @@ func TestChannelManager_Close_BeforeMPTBlock(t *testing.T) {
 
 	require.NoError(m.AddL2Block(zktBlock1), "adding 1st L2 block")
 	require.NoError(m.AddL2Block(zktBlock2), "adding 2nd L2 block")
-	require.NoError(m.AddL2Block(mptBlock), "adding 3nd L2 block")
+	require.NoError(m.AddL2Block(mptBlock), "adding 3rd L2 block")
```
0xbenyun, line 529: How about renaming the test function to something like TestChannelManager_CheckFullErrorsOnMPTHardFork?

sm-stack (approved):

seolaoh (approved):

0xbenyun (approved):

kangsorang (approved):

PR #421, opened 2025-01-23
chore: Fix incorrect type for Votes Update governancetoken.go

mdqst:
# Description

I noticed a type mismatch in the `ERC20VotesCheckpoint` struct. The `Votes` field was defined as `*big.Int`, but according to the ABI contract, it should be `uint224`. This discrepancy could cause issues when interacting with the contract.  

I’ve updated the struct to use `uint224` for the `Votes` field to align with the ABI specification. This should resolve any potential errors related to type mismatches.


Pangssu:
Hi @mdqst, thank you for your contribution. 
This file is auto-generated code. It was generated by `abigen` of ethereum, so we can't apply this change.

PR #428, opened 2025-01-25
chore: fix some comments

jiangmencity:
# Description

fix some comments


Pangssu:
We are currently upstreaming to the latest version of Optimism, and it will be applied once the process is complete.

Pangssu (approved):

kangsorang:
Thanks for this. This is a reasonable fix, but since the Kroma repo is upstream of the optimism repo, I think it would be better to target the optimism repo first for this fix.

PR #431, opened 2025-01-26
Add nil check for storage values in SetStorage function

Daulox92:
Changes:
- Added nil check validation in kroma-chain-ops/state/state.go
- Added error message for nil values parameter

Why:
- Prevents potential panic from nil values
- Provides early error detection with clear message
- Follows Go best practices for defensive programming
- No functional changes, only validation improvement

Testing:
- Verified with nil values for error handling
- Verified with valid values for normal operation


Pangssu:
Looking at where this function is used, it is only called when `storageConfig` exists.

See https://github.com/kroma-network/kroma/blob/dev/kroma-chain-ops/genesis/setters.go#L94

PR #432, opened 2025-01-27
docs: fix grammar and documentation issues

Hopium21:
Changes:

1. kroma-chain-ops/cmd/check-kroma-mgt/main.go:
"fee distribution unexpectedly"
"fee distribution unexpected"
Reason: Fixed grammatical error by using correct adjective form.

2. kroma-chain-ops/crossdomain/message_test.go:
// TestEncode tests the hash of a CrossDomainMessage.
// TestHash tests the hash of a CrossDomainMessage.
Reason: Fixed incorrect function name in comment to match actual function.


sm-stack:
The latter one is from optimism repo, and Kroma is following optimism's code. I'd recommend to contribute to that repo, which will be applied to Kroma too. @Hopium21  Could you drop the latter commit?

PR #440, opened 2025-03-25
feat: challenge protocol update

0xbenyun:
# Description

To achieve stage 1, this PR modifies the bisection process and updates the criteria for output finalization. Additionally, this PR only includes changes to the contract components. The validator will be developed in a separate PR.


kangsorang:
I wish you could update the storageLayout as well~!

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -1,15 +1,15 @@
...
+import {IERC721Receiver} from "@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol";
 
-import { Uint128Math } from "../libraries/Uint128Math.sol";
-import { ISemver } from "../universal/ISemver.sol";
-import { IAssetManager } from "./interfaces/IAssetManager.sol";
-import { IValidatorManager } from "./interfaces/IValidatorManager.sol";
+import {Uint128Math} from "../libraries/Uint128Math.sol";
+import {ISemver} from "../universal/ISemver.sol";
+import {IAssetManager} from "./interfaces/IAssetManager.sol";
+import {IValidatorManager} from "./interfaces/IValidatorManager.sol";
```
Pangssu: Please check your formatter configuration. These should be like this.
```
import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```

packages/contracts/contracts/L1/AssetManager.sol
```diff
@@ -679,9 +679,9 @@ contract AssetManager is ISemver, IERC721Receiver, IAssetManager {
     ) internal view returns (uint128) {
         return
             assets.mulDiv(
-                _totalKroShares(validator) + DECIMAL_OFFSET,
-                totalKroAssets(validator) + 1
-            );
+            _totalKroShares(validator) + DECIMAL_OFFSET,
+            totalKroAssets(validator) + 1
+        );
```
Pangssu: Please check your formatter configuration.

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -49,61 +50,87 @@ contract Colosseum is Initializable, ISemver {
...
+    mapping(uint256 => Types.Assertion) public assertions;
 
     /**
      * @notice A mapping of the challenge.
      */
     mapping(uint256 => mapping(address => Types.Challenge)) public challenges;
 
     /**
-     * @notice A mapping indicating whether a public input is verified or not.
+   * @notice A modifier that only allows L2OutputOracle contract to call.
```
Pangssu:
```suggestion
     * @notice A modifier that only allows L2OutputOracle contract to call.
```

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -198,7 +225,7 @@ contract Colosseum is Initializable, ISemver {
      * @param challenger  Address of the challenger.
      * @param timestamp   The timestamp when deleted.
      */
-    event ChallengerTimedOut(
+    event ChallengerHasTimedOut(
```
Pangssu: Do not change the event name. It makes difference of topic hash.

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -260,49 +287,69 @@ contract Colosseum is Initializable, ISemver {
...
 
     /**
-     * @notice Reverts when challenge cannot be cancelled.
+     * @notice Reverts when a challenge cannot be created or progressed.
      */
-    error CannotCancelChallenge();
+    error NotChallengeable();
+
+    /**
+     * @notice Reverts when ch
```
Pangssu:
```suggestion
     * @notice Reverts when the challenge has been already created.
```

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -426,106 +496,107 @@ contract Colosseum is Initializable, ISemver {
...
+        // If assertion doesn't exist
+        if (assertion.assertedAt == 0) {
+            revert InvalidOutputGiven();
+        }
+
+        if (assertion.status != Types.AssertionStatus.IN_PROGRESS) {
+            revert NotChallengeable();
+        }
+
+        Types.Challenge  storage challenge = challenges[_outputIndex][_challenger];
```
Pangssu:
```suggestion
        Types.Challenge storage challenge = challenges[_outputIndex][_challenger];
```

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -544,6 +615,35 @@ contract Colosseum is Initializable, ISemver {
...
+    * @notice Accepts the assertion.
+     *         Reverts if is not possible to accept the assertion.
+     *
+     * @param _outputIndex Index of the L2 checkpoint output.
+     */
+    function acceptAssertion(uint256 _outputIndex) external {
+        Types.Assertion storage assertion = assertions[_outputIndex];
+        if (!_acceptAssertionIfNoChallengeAndTimedOut(assertion)) {
+            revert AssertionNotAcceptable();
+        }
```
Pangssu: If `_acceptAssertionIfNoChallengeAndTimedOut` is not used elsewhere within the contract,
then there's no need to separate it as an internal function.

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -646,108 +746,50 @@ contract Colosseum is Initializable, ISemver {
...
-        bytes32 publicInputHash;
-        if (_isZkVm) {
-            publicInputHash = ZK_PROOF_VERIFIER.verifyZkVmProof(
-                _zkVmProof,
-                srcSegment,
-                dstSegment,
-                challenge.l1Head
-            );
+
+        bytes32 srcOutput = bytes32(_zkVmProof.publicValues[8 : 40]);
```
Pangssu: Please describe the reason for slicing from 8 to 40.

packages/contracts/contracts/L1/KromaPortal.sol
```diff
@@ -147,6 +154,11 @@ contract KromaPortal is Initializable, ResourceMetering, ISemver {
      */
     event Unpaused(address account);
 
+    /**
+     * @notice Reverts when finalizeWithdrawalTransaction is called on a deleted output.
+     */
+    error DeletedOutput();
```
Pangssu: I think it would be better to add a prefix like 'Err'.
```suggestion
    error ErrDeletedOutput();
```

packages/contracts/contracts/L1/KromaPortal.sol
```diff
@@ -375,6 +385,7 @@ contract KromaPortal is Initializable, ResourceMetering, ISemver {
             "KromaPortal: withdrawal timestamp less than L2 Oracle starting timestamp"
         );
 
+        // TODO: why is this necessary?
```
Pangssu: Because it is part of two step withdrawal process. 
See https://blog.oplabs.co/two-step-withdrawals/

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -116,8 +117,8 @@ contract L2OutputOracle is Initializable, ISemver {
      * @param _finalizationPeriodSeconds Output finalization time in seconds.
      */
     constructor(
-        ValidatorPool _validatorPool,
-        IValidatorManager _validatorManager,
+        address _validatorPool,
```
Pangssu, line 120: We don't need validator pool contract anymore.

```suggestion
```

packages/contracts/contracts/libraries/Types.sol
```diff
@@ -146,27 +146,67 @@ library Types {
...
      * @custom:field asserter   Address of the asserter.
-     * @custom:field challenger Address of the challenger.
-     * @custom:field segments   Array of the segment.
-     * @custom:field segStart   The L2 block number of the first segment.
-     * @custom:field segSize    The number of L2 blocks.
-     * @custom:field l1Head     Parent L1 block hash at the challenge creation time.
+     * @custom:field assertedAt timestamp of Assertion.
+     * @custom:field acceptedAt
+     * @custom:field rejectedAt
+     * @custom:field status
```
Pangssu: Please describe them

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -453,17 +456,18 @@ contract L2OutputOracle is Initializable, ISemver {
...
-     */
-    function finalizedAt(uint256 _outputIndex) public view returns (uint256) {
-        return l2Outputs[_outputIndex].timestamp + FINALIZATION_PERIOD_SECONDS;
+        Types.Assertion memory assertion = IColosseum(COLOSSEUM).getAssertion(_outputIndex);
+        if ((assertion.assertedAt != 0 && assertion.status == Types.AssertionStatus.IN_PROGRESS && assertion.numberOfChallenges == 0 && block.timestamp - assertion.assertedAt > IColosseum(COLOSSEUM).GUARDIAN_PERIOD() + IColosseum(COLOSSEUM).MAX_CLOCK_DURATION())
+            || (assertion.acceptedAt != 0 && block.timestamp - assertion.acceptedAt > IColosseum(COLOSSEUM).GUARDIAN_PERIOD())
+        ) {
+            return true;
+        }
+        return false;
```
Pangssu: That’s a complex conditional statement. I think it could be improved like this.

```suggestion
    IColosseum colosseum = IColosseum(COLOSEUM);
    Types.Assertion memory assertion = colosseum.getAssertion(_outputIndex);
    if (assertion.assertedAt == 0) {
        return false;
    }
    
    if (assertion.status == Types.AssertionStatus.IN_PROGRESS) {
        if (assertion.numberOfChallenges > 0) {
            return false;
        }
        if (block.timestamp < assertion.assertedAt + colosseum.GUARDIAN_PERIOD() + colosseum.MAX_CLOCK_DURATION()) {
            return false;
        }
    } else if (assertion.status == Types.AssertionStatus.ACCEPTED) {
        if (block.timestamp < assertion.acceptedAt + colosseum.GUARDIAN_PERIOD()) {
            return false;
        }
    } else {
        return false;
    }
    return true;
```

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -617,6 +616,7 @@ contract ValidatorManager is ISemver, IValidatorManager {
                     ++outputIndex;
                     ++finalizedOutputNum;
                 }
+
```
Pangssu: remove unnecessary new line

packages/contracts/contracts/L1/ValidatorManager.sol
```diff
@@ -680,7 +680,7 @@ contract ValidatorManager is ISemver, IValidatorManager {
             uint128 validatorKro = ASSET_MANAGER.totalValidatorKro(validator);
             uint128 totalKro = ASSET_MANAGER.totalKroAssets(validator);
             uint128 validatorBaseReward = baseReward.mulDiv(validatorKro, totalKro + validatorKro);
-            // Exclude the base reward for the validator from total base reward given to KRO delegators.
+        // Exclude the base reward for the validator from total base reward given to KRO delegators.
```
Pangssu:
```suggestion
            // Exclude the base reward for the validator from total base reward given to KRO delegators.
```

Pangssu:
Please reformat all changes after properly configuring your formatter.

packages/contracts/contracts/L1/AssetManager.sol
0xbenyun: Updated as suggested

packages/contracts/contracts/L1/AssetManager.sol
0xbenyun: Updated as suggested

packages/contracts/contracts/L1/Colosseum.sol
0xbenyun: Updated as suggested

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -316,51 +362,60 @@ contract Colosseum is Initializable, ISemver {
...
         ZK_PROOF_VERIFIER = _zkProofVerifier;
-        CREATION_PERIOD_SECONDS = _creationPeriodSeconds;
-        BISECTION_TIMEOUT = _bisectionTimeout;
-        PROVING_TIMEOUT = _provingTimeout;
         L2_ORACLE_SUBMISSION_INTERVAL = _submissionInterval;
         SECURITY_COUNCIL = _securityCouncil;
-        initialize(_segmentsLengths);
+        GUARDIAN_PERIOD = _guardianPeriod;
+        MAX_CLOCK_DURATION = _maxClockDuration;
+        initialize();
```
kangsorang: How about this?
```suggestion
        _disableInitializers();
```

packages/contracts/contracts/L1/Colosseum.sol
0xbenyun: Updated as suggested

packages/contracts/contracts/L1/Colosseum.sol
0xbenyun: Updated as suggested

packages/contracts/contracts/L1/Colosseum.sol
0xbenyun: Updated as suggested

packages/contracts/contracts/L1/Colosseum.sol
0xbenyun: Updated as suggested

packages/contracts/contracts/L1/KromaPortal.sol
0xbenyun: The error was unnecessary, so I removed it. Thank you for your feedback

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -777,7 +800,7 @@ contract Colosseum is Initializable, ISemver {
 
         // Switch validator system after validator pool contract terminated.
         if (L2_ORACLE.VALIDATOR_POOL().isTerminated(_outputIndex)) {
```
kangsorang, line 830: It would be nice to remove the Validator System V1 code here as well.

packages/contracts/contracts/L1/KromaPortal.sol
```diff
@@ -177,13 +184,15 @@ contract KromaPortal is Initializable, ResourceMetering, ISemver {
...
-        ZKMerkleTrie _zkMerkleTrie
+        ZKMerkleTrie _zkMerkleTrie,
+        address _colosseum
     ) {
         L2_ORACLE = _l2Oracle;
         VALIDATOR_POOL = _validatorPool;
         GUARDIAN = _guardian;
         SYSTEM_CONFIG = _config;
         ZK_MERKLE_TRIE = _zkMerkleTrie;
+        COLOSSEUM = IColosseum(_colosseum);
```
kangsorang: I have a question. 
Do we need the colosseum address in the portal?

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -369,13 +424,11 @@ contract Colosseum is Initializable, ISemver {
...
      */
     function createChallenge(
         uint256 _outputIndex,
         bytes32 _l1BlockHash,
-        uint256 _l1BlockNumber,
-        bytes32[] calldata _segments
+        uint256 _l1BlockNumber
     ) external {
         if (_outputIndex == 0) revert NotAllowedGenesisOutput();
         // Switch validator system after validator pool contract terminated.
```
kangsorang, line 433: I think the ValidatorPool stuff down here should be removed.

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -49,61 +49,87 @@ contract Colosseum is Initializable, ISemver {
...
+    mapping(bytes32 => bool) public verifiedPublicInputs;
+
+    /**
+     * @notice A mapping of deleted output index to the deleted output.
+     */
+    mapping(uint256 => Types.CheckpointOutput) public deletedOutputs;
 
     /**
-     * @notice Length of segment array for each turn.
+     * @notice A mapping of the assertion.
```
kangsorang: It would be helpful to include a more detailed comment for the mapping field, explaining both the key and value types.
```suggestion
     * @notice Maps each output index to its corresponding assertion object.
```

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -316,51 +362,60 @@ contract Colosseum is Initializable, ISemver {
...
+        Types.Assertion storage assertion = assertions[_outputIndex];
+        if (assertion.asserter != address(0)) {
+            revert AssertionAlreadyCreated();
+        }
+
+        assertion.startL2BlockNumber = start;
+        assertion.startOutputRoot = finalizedOutput.outputRoot;
+        assertion.asserter = asserter;
+        assertion.assertedAt = block.timestamp;
+
```
kangsorang: Although IN_PROGRESS is the default value for the enum, it would be clearer to set it explicitly in the code.
Also, it seems that status transitions for other AssertionStatus values are missing — it might be good to check where and how those updates are handled.
```suggestion
assertion.status = Types.AssertionStatus.IN_PROGRESS;
```

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -389,17 +442,21 @@ contract Colosseum is Initializable, ISemver {
...
-            if (_challengeStatus(challenge) != ChallengeStatus.CHALLENGER_TIMEOUT)
+            if (_challengeStatus(challenge) != ChallengeStatus.CHALLENGER_TIMEOUT) {
                 revert ImproperChallengeStatus();
-
+            }
             _challengerTimeout(_outputIndex, msg.sender);
         }
 
-        Types.CheckpointOutput memory targetOutput = L2_ORACLE.getL2Output(_outputIndex);
+        Types.Assertion storage assertion = assertions[_outputIndex];
```
kangsorang: It would be good to add handling for the case where the given assertion does not exist.

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -389,17 +442,21 @@ contract Colosseum is Initializable, ISemver {
...
 
-        if (targetOutput.timestamp + CREATION_PERIOD_SECONDS < block.timestamp)
-            revert CreationPeriodPassed();
+        if (assertion.status != Types.AssertionStatus.IN_PROGRESS) {
+            revert NotChallengeable();
```
kangsorang, line 456: I couldn’t find the code that updates the status to ACCEPTED, REJECTED, or RESTORED.
Is it possible that I missed it, or has it not been implemented yet?

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -49,61 +49,87 @@ contract Colosseum is Initializable, ISemver {
         READY_TO_PROVE
     }
 
+    /**
+     * @notice A duration for asserter(or challenger) timeout.
+     */
+    uint256 public immutable MAX_CLOCK_DURATION;
```
kangsorang: It might be helpful to include the time unit in the constant name for better clarity.
```suggestion
    uint256 public immutable MAX_CLOCK_DURATION_SECONDS;
```

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -464,68 +531,67 @@ contract Colosseum is Initializable, ISemver {
...
+        if (status == ChallengeStatus.ASSERTER_TURN) {
+            if (msg.sender != challenge.asserter) {
+                revert NotAllowedCaller();
+            }
+            challenge.asserterTimeLeft -= elapsed;
+        } else if (status == ChallengeStatus.CHALLENGER_TURN) {
+            if (msg.sender != challenge.challenger) {
+                revert NotAllowedCaller();
+            }
+            challenge.challengerTimeLeft -= elapsed;
```
kangsorang, line 555: Maybe it would be clearer to update both asserterTimeLeft and challengerTimeLeft based on the elapsed time, regardless of whose turn it is.
```suggestion
        if (status == ChallengeStatus.ASSERTER_TURN) {
            if (msg.sender != challenge.asserter) {
                revert NotAllowedCaller();
            }
        } else if (status == ChallengeStatus.CHALLENGER_TURN) {
            if (msg.sender != challenge.challenger) {
                revert NotAllowedCaller();
            }
        ....
        challenge.asserterTimeLeft -= elapsed;
        challenge.challengerTimeLeft -= elapsed;
```

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -464,68 +531,67 @@ contract Colosseum is Initializable, ISemver {
...
         challenge.turn = newTurn;
-        _updateTimeout(challenge);
+        challenge.segment.output = _output;
+        challenge.updatedAt = block.timestamp;
+        challenge.segment.pos = _pos;
 
-        emit Bisected(_outputIndex, _challenger, newTurn, block.timestamp);
-
-        if (!_isAbleToBisect(challenge)) {
+        if (_challengeStatus(challenge) == ChallengeStatus.READY_TO_PROVE) {
```
kangsorang, line 586: Currently, it seems difficult to track changes to the ChallengeStatus.
Although the status is internally updated when functions like bisect are called,
the only way to observe the status at a given point in time is by calling _challengeStatus directly.
Would it make sense to add an event like ChallengeStatusChanged to make these transitions easier to track externally? e.g.,
```
event ChallengeStatusChanged(
    uint256 indexed outputIndex,
    address indexed challenger,
    ChallengeStatus previousStatus,
    ChallengeStatus newStatus,
    uint256 timestamp
);

ChallengeStatus prevStatus = status;
ChallengeStatus newStatus = _challengeStatus(challenge);
if (prevStatus != newStatus) {
    emit ChallengeStatusChanged(_outputIndex, _challenger, prevStatus, newStatus, block.timestamp);
}
```

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -544,6 +610,23 @@ contract Colosseum is Initializable, ISemver {
...
+     */
+    function acceptAssertion(uint256 _outputIndex) external {
+        Types.Assertion storage assertion = assertions[_outputIndex];
+        if (
+            assertion.numberOfChallenges > 0 ||
+            block.timestamp - assertion.assertedAt < MAX_CLOCK_DURATION
+        ) {
+            revert AssertionNotAcceptable();
+        }
+
```
kangsorang, line 640: Should we add a check to ensure the assertion is still IN_PROGRESS before accepting it?
```suggestion
if (assertion.status != Types.AssertionStatus.IN_PROGRESS) {
    revert AssertionNotAcceptable();
}
```

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -646,108 +729,48 @@ contract Colosseum is Initializable, ISemver {
...
-            publicInputHash = ZK_PROOF_VERIFIER.verifyZkVmProof(
-                _zkVmProof,
-                srcSegment,
-                dstSegment,
-                challenge.l1Head
-            );
+        // Slice from index 8 to 40 to extract the srcOutputRoot,
+        // as publicValues contains concatenated bytes32 values.
+        // Each bytes32 value occupies 32 bytes, so this range corresponds to the first public input.
+        bytes32 srcOutput = bytes32(_zkVmProof.publicValues[8:40]);
```
kangsorang, line 786: If we’re only keeping part of the segment after the split,
maybe the client could just pass in the relevant half, rather than calculating and sending pos?

packages/contracts/contracts/libraries/Types.sol
```diff
@@ -149,26 +149,89 @@ library Types {
...
+     * @notice Struct representing a assertion.
+     *
+     * @custom:field startL2BlockNumber  start l2 block number for bisection.
+     * @custom:field startOutputRoot   output.
+     * @custom:field asserter   Address of the asserter.
+     * @custom:field assertedAt Timestamp when the assertion was created
+     * @custom:field acceptedAt Timestamp when the assertion was accepted.
+     * @custom:field rejectedAt Timestamp when the assertion was rejected.
+     * @custom:field status Current status of the assertion.
+     * @custom:field numberOfChallenges Number of challenges raised against the assertion.
```
Pangssu:
```suggestion
     * @custom:field startL2BlockNumber Starting L2 block number for bisection.
     * @custom:field startOutputRoot    Starting L2 output root for bisection.
     * @custom:field asserter           Address of the asserter.
     * @custom:field assertedAt         Timestamp when the assertion was created
     * @custom:field acceptedAt         Timestamp when the assertion was accepted.
     * @custom:field rejectedAt         Timestamp when the assertion was rejected.
     * @custom:field status Current     Status of the assertion.
     * @custom:field numberOfChallenges Number of challenges raised against the assertion.
```

packages/contracts/contracts/libraries/Types.sol
```diff
@@ -149,26 +149,89 @@ library Types {
...
+        uint256 assertedAt;
+        uint256 acceptedAt;
+        uint256 rejectedAt;
+        AssertionStatus status;
+        uint256 numberOfChallenges;
+    }
+
+    /**
+   * @notice Enum of the Assertion status.
+     *
```
Pangssu:
```suggestion
    /**
     * @notice Enum of the Assertion status.
     *
```

packages/contracts/contracts/L1/L2OutputOracle.sol
0xbenyun: This is intuitive code. I have modified it based on this. Thank you.

packages/contracts/contracts/L1/KromaPortal.sol
0xbenyun: Good point. It is unnecessary, and I will remove it.

packages/contracts/contracts/L1/Colosseum.sol
0xbenyun: Thank you. While analyzing the part you pointed out, I realized that the initialize() call itself should be removed. Regarding your suggestion, reinitialization might be needed if the Colosseum contract undergoes an upgrade in the future. So, I will simply remove the initialize code.

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -260,49 +286,69 @@ contract Colosseum is Initializable, ISemver {
...
      */
-    error AlreadyVerifiedPublicInput();
+    error AssertionAlreadyCreated();
 
     /**
-     * @notice Reverts when the public input hash is invalid.
+     * @notice Reverts when an asserter has timed out
      */
-    error InvalidPublicInputHash();
+    error AsserterTimeoutError();
```
dongchangYoo: How about removing the suffix ‘Error’ for consistency?

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -260,49 +286,69 @@ contract Colosseum is Initializable, ISemver {
...
 
     /**
-     * @notice Reverts when turn is invalid.
+     * @notice Reverts when a challenger has timed out
      */
-    error InvalidTurn();
+    error ChallengerTimeoutError();
```
dongchangYoo: How about removing the suffix ‘Error’ for consistency?

packages/contracts/contracts/test/AssetManager.t.sol
```diff
@@ -100,7 +100,7 @@ contract AssetManagerTest is ValidatorSystemUpgrade_Initializer {
         for (uint256 i = oracle.nextOutputIndex(); i <= terminateOutputIndex; i++) {
             _submitOutputRoot(pool.nextValidator());
         }
-        vm.warp(oracle.finalizedAt(terminateOutputIndex));
+        //        vm.warp(oracle.finalizedAt(terminateOutputIndex));
```
dongchangYoo: Is it should removed? or did you leave this line on purpose?

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -316,51 +362,60 @@ contract Colosseum is Initializable, ISemver {
...
-            unchecked {
-                ++i;
-            }
+        uint256 finalizedOutputIndex = L2_ORACLE.nextFinalizeOutputIndex();
+        if (finalizedOutputIndex > 0) {
+            finalizedOutputIndex--;
         }
 
-        if (sum != L2_ORACLE_SUBMISSION_INTERVAL) revert InvalidSegmentsLength();
+        Types.CheckpointOutput memory finalizedOutput = L2_ORACLE.getL2Output(finalizedOutputIndex);
```
dongchangYoo: It seems that the few lines above, including this one, are actually logic that should be executed in L2OutputOracle. If you need a pattern where data is fetched from L2Oracle and then passed back to L2Oracle, how about implementing this logic in L2Oracle?

I recommends to add `getFinalizedL2Output` to L2Oracle contract. Then, the logic of `validatorManager::_updatePriorityValidator` can also be simplified.

packages/contracts/contracts/L1/Colosseum.sol
0xbenyun: The logic for handling that case is already implemented. The relevant code is as follows:

`if (assertion.asserter != address(0)) {
    revert AssertionAlreadyCreated();
}`

packages/contracts/contracts/L1/Colosseum.sol
0xbenyun: I missed it. thx!

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -316,51 +362,60 @@ contract Colosseum is Initializable, ISemver {
...
+
+        uint256 start = finalizedOutput.l2BlockNumber;
+        Types.Assertion storage assertion = assertions[_outputIndex];
+        if (assertion.asserter != address(0)) {
+            revert AssertionAlreadyCreated();
+        }
+
+        assertion.startL2BlockNumber = start;
+        assertion.startOutputRoot = finalizedOutput.outputRoot;
```
dongchangYoo: These two variables come from a `finalizedOutput`. Since the `finalizedOutput` will not change, it is recommended to retrieve them only when needed rather than storing two slots every time. Storing only the `outputIndex` will save gas costs for Validators.

dongchangYoo:
The `AssertionCreated` and `OutputSubmitted` events always occur together. How about merging them into a single event?

packages/contracts/contracts/L1/Colosseum.sol
kangsorang: In the upstream Optimism repo, the contracts have been updated to use the Initializable pattern. Thought it might be helpful to keep this in mind.
[https://github.com/kroma-network/optimism/blob/feat/apply-kroma-v2.1.0/packages/contracts-bedrock/src/L1/Colosseum.sol#L199-L221](url)

packages/contracts/contracts/L1/Colosseum.sol
kangsorang: In this case, checking if (assertion.status != Types.AssertionStatus.IN_PROGRESS) would cover that scenario. 
Got it.

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -270,6 +270,9 @@ contract L2OutputOracle is Initializable, ISemver {
                 l2BlockNumber: uint128(_l2BlockNumber)
             })
         );
+        if (outputIndex > 0) {
```
dongchangYoo, line 273: The createAssertion function checks if outputIndex is zero, so this `if` is not necessary.

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -453,17 +456,37 @@ contract L2OutputOracle is Initializable, ISemver {
...
-    function finalizedAt(uint256 _outputIndex) public view returns (uint256) {
-        return l2Outputs[_outputIndex].timestamp + FINALIZATION_PERIOD_SECONDS;
+        IColosseum colosseum = IColosseum(COLOSSEUM);
+        Types.Assertion memory assertion = colosseum.getAssertion(_outputIndex);
+        if (assertion.assertedAt == 0) {
+            return false;
+        }
+
+        if (assertion.status == Types.AssertionStatus.RESTORED) {
+            return true; // RESTORED assertions are considered finalized due to social resolution (guardian mechanism).
```
dongchangYoo: As far as i know, a RESTORED assertion that is less than 7 days old should be considered not finalized, right?

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -453,17 +456,37 @@ contract L2OutputOracle is Initializable, ISemver {
...
+        } else if (assertion.status == Types.AssertionStatus.ACCEPTED) {
+            if (block.timestamp <= assertion.acceptedAt + colosseum.GUARDIAN_PERIOD()) {
+                return false;
+            }
```
dongchangYoo:
``` suggestion
            return block.timestamp > (assertion.acceptedAt + colosseum.GUARDIAN_PERIOD());
```

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -453,17 +456,37 @@ contract L2OutputOracle is Initializable, ISemver {
...
+        } else if (assertion.status == Types.AssertionStatus.IN_PROGRESS) {
+            if (assertion.numberOfChallenges > 0) {
+                return false;
+            }
+            if (
+                block.timestamp <=
+                assertion.assertedAt + colosseum.GUARDIAN_PERIOD() + colosseum.MAX_CLOCK_DURATION()
+            ) {
+                return false;
+            }
```
dongchangYoo:
``` suggestion
            return (assertion.numberOfChallenges == 0) &&
               (block.timestamp > (assertion.assertedAt + colosseum.GUARDIAN_PERIOD() + colosseum.MAX_CLOCK_DURATION()));
```

packages/contracts/contracts/L1/L2OutputOracle.sol
```diff
@@ -453,17 +456,37 @@ contract L2OutputOracle is Initializable, ISemver {
...
+        } else {
+            return false;
+        }
+
+        return true;
```
dongchangYoo: This is an unreachable line. In Solidity, is it considered best practice to include a return statement even in such cases?

packages/contracts/contracts/L1/Colosseum.sol
0xbenyun: Since it follows the chess clock model, for example, during the Asserter’s turn, the Challenger’s time should not be consumed, and only the Asserter’s time should be decremented. Let me know if you have any thoughts.

packages/contracts/contracts/L1/L2OutputOracle.sol
0xbenyun: I added this condition because when outputIndex is 0, the createAssertion function is not called, but the output still needs to be submitted.
(The output for outputIndex 0 is excluded from assertion & challenge and is considered a finalized output.)

packages/contracts/contracts/L1/L2OutputOracle.sol
0xbenyun: This is a reachable line. For example, if none of the if conditions in the above else if statements are met to return false, execution will reach this line.

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -409,15 +468,32 @@ contract Colosseum is Initializable, ISemver {
...
+        challenge.asserter = assertion.asserter;
+        challenge.turn = TURN_INIT;
 
-        // If the previous output has been deleted, the first segment will not be compared with the previous output.
-        if (prevOutput.outputRoot == DELETED_OUTPUT_ROOT) {
-            _validateSegments(TURN_INIT, _segments[0], targetOutput.outputRoot, _segments);
-        } else {
-            _validateSegments(TURN_INIT, prevOutput.outputRoot, targetOutput.outputRoot, _segments);
+        uint256 elapsed = block.timestamp - assertion.assertedAt;
+        if (elapsed >= MAX_CLOCK_DURATION_SECONDS) {
```
Pangssu:
```suggestion
        if (elapsed > MAX_CLOCK_DURATION_SECONDS) {
```
is better

packages/contracts/contracts/L1/Colosseum.sol
```diff
@@ -426,106 +502,113 @@ contract Colosseum is Initializable, ISemver {
             L2_ORACLE.VALIDATOR_POOL().addPendingBond(_outputIndex, msg.sender);
         }
```
Pangssu, line 502: Remove these lines L421-L427 except L424

kangsorang:
Coderabbit is going wild… Is there anything we can do to stop it?

packages/contracts/contracts/L1/Colosseum.sol
0xbenyun: I will make sure to reflect this when migrating to the Optimism repository.

packages/contracts/contracts/L1/Colosseum.sol
0xbenyun: I haven’t yet identified a need for the event based on changes to the Challenge Status. How about we implement it later if the need arises?

packages/contracts/contracts/libraries/Types.sol
```diff
@@ -198,15 +198,16 @@ library Types {
...
      */
     struct Assertion {
         uint256 latestFinalizedOutputIndex;
         address asserter;
         uint256 assertedAt;
         uint256 acceptedAt;
         uint256 rejectedAt;
-        AssertionStatus status;
         uint256 numberOfChallenges;
+        bool isEnforced;
```
Pangssu: Pack these value to reduce gas usage.

packages/contracts/contracts/libraries/Types.sol
```diff
@@ -221,13 +222,13 @@ library Types {
      *  1) IN_PROGRESS      → createAssertion()
      *  2) ACCEPTED      → when challenger's timer has expired
      *  3) REJECTED      → when proveFault() succeeds or asserter timeout
-     *  4) RESTORED    → when deleted output is restored by SC
+     *  4) ENFORCED    → when a assertion is enforced by SC
```
Pangssu:
```suggestion
     *  1) IN_PROGRESS → createAssertion()
     *  2) ACCEPTED    → when challenger's timer has expired
     *  3) REJECTED    → when proveFault() succeeds or asserter timeout
     *  4) RESTORED    → when deleted output is restored by Security Council
```

packages/contracts/contracts/L1/Colosseum.sol
0xbenyun: I think the current structure should be maintained because there is a verification logic in the ZKProofVerifier using the corresponding publicValues.

0xbenyun:
@coderabbitai pause

packages/contracts/contracts/L1/Colosseum.sol
Pangssu: Since the ChallengeStatusChanged event was originally being emitted, it would be better to leave it as is.

Pangssu:
I disabled a coderabbit in this repository.

packages/contracts/contracts/L1/Colosseum.sol
kangsorang: This process requires a two-step approach due to contract dependencies.
In the first step, conditional logic is added to ZKProofVerifier based on the length of public values.
After Colosseum V2 is upgraded with improved public values, the conditional logic in ZKProofVerifier can be safely removed.
However, since it’s not strictly required functionally, it can be applied optionally.

packages/contracts/contracts/L1/Colosseum.sol
kangsorang: > Since the ChallengeStatusChanged event was originally being emitted, it would be better to leave it as is.

You’re referring to events like ChallengeCanceled, ChallengeCreated, and ChallengeDismissed, right?
Since there are already established events that serve those purposes, it wouldn’t be ideal to change them easily.
I was thinking that instead of subscribing to each specific event, handling things more broadly through a single ChallengeStatusChanged event could offer several advantages.

packages/contracts/contracts/L1/Colosseum.sol
kangsorang: On second thought, this might be a bit outside the scope of the challenge protocol.
If we do decide to move forward with it, it would probably be better handled in a separate PR.

0xbenyun: That’s a great suggestion. Let’s handle it in a separate PR.

packages/contracts/contracts/L1/Colosseum.sol
0xbenyun: It makes sense to handle this in a separate PR as you suggested.

packages/contracts/contracts/L1/Colosseum.sol
kangsorang: It would be better to handle this refactoring in a separate PR.

packages/contracts/contracts/test/AssetManager.t.sol
0xbenyun: I deleted this source code since it no longer gets finalized simply after a certain period of time.

packages/contracts/contracts/L1/Colosseum.sol
0xbenyun: As we discussed, it has been decided to handle this in a future PR.

Pangssu (approved):

kangsorang (approved):
LGTM

